PR #10, opened 2026-03-06
fix(browser-preview): suppress turnstile error callbacks and improve cf challenge detection

FayruzAldi:


ArgaFairuz (changes requested):
It’s perfect, but it causes a memory issue because you removed puppeteer-cluster without reconsidering its role. Maybe it could be adapted to work with puppeteer-extra if possible. Thanks!

PR #11, opened 2026-03-06
fix(browser-preview): virtual cursor dind'nt fanished instantly after task completion

FayruzAldi:


ArgaFairuz (changes requested):
It's perfect, but it brings back fix #2, the virtual cursor disappears every time MCP finishes moving it.

ArgaFairuz:
Resolved in #50

PR #223, opened 2026-05-02
fix(db-client): scope connections by owner and redact sensitive credentials

DeryFerd:
This PR adds ownership enforcement for db-client connections and prevents credential leakage in API responses.

## Why?
DB client connections were globally accessible and returned sensitive credential fields in list/get responses.

## Changes
- Added migration `035_add_owner_to_db_client_connections`:
- introduces `owner_user_id`
- adds index for owner lookups
- Added ownership-aware query methods for db-client connections.
- Enforced per-user/admin access checks across all `db-client:*` WebSocket handlers.
- Redacted sensitive fields from response payloads:
- `password`
- `ssh.password`
- `ssh.privateKey`
- `ssh.passphrase`
- Wired query history writes with current user context.
- Preserved existing DB and SSH secrets when editing redacted connection forms.

## Security impact
- Prevents unauthorized users from using or modifying other users' saved DB connections.
- Prevents accidental exposure of stored credentials through API responses.

## Legacy data note
Existing db-client connections created before migration `035_add_owner_to_db_client_connections` have `owner_user_id = NULL`. After upgrading, those legacy connections remain visible to admins only because there is no reliable historical owner signal. Operators should ask users to recreate legacy connections or manually assign ownership via SQL when appropriate.

## Note
Admins still retain global access by design; regular users are restricted to their own connections.

### DIFF
--- a/backend/database/migrations/035_add_owner_to_db_client_connections.ts
+++ b/backend/database/migrations/035_add_owner_to_db_client_connections.ts
@@ -1,0 +1,34 @@
+import type { DatabaseConnection } from '$shared/types/database/connection';
+import { debug } from '$shared/utils/logger';
+
+export const description = 'Add owner_user_id to db_client_connections table';
+
+export const up = (db: DatabaseConnection): void => {
+	debug.log('migration', 'Adding owner_user_id to db_client_connections...');
+
+	// SQLite does not support IF NOT EXISTS for ADD COLUMN, so we ignore duplicate-column errors.
+	try {
+		db.exec(`
+			ALTER TABLE db_client_connections
+			ADD COLUMN owner_user_id TEXT
+		`);
+	} catch (error) {
+		const message = error instanceof Error ? error.message : String(error);
+		if (!message.toLowerCase().includes('duplicate column name')) {
+			throw error;
+		}
+	}
+
+	db.exec(`
+		CREATE INDEX IF NOT EXISTS idx_db_client_conn_owner
+		ON db_client_connections(owner_user_id)
+	`);
+
+	debug.log('migration', 'owner_user_id migration completed');
+};
+
+export const down = (db: DatabaseConnection): void => {
+	// SQLite cannot drop columns without table rebuild; keep rollback non-destructive.
+	debug.log('migration', 'Dropping idx_db_client_conn_owner index...');
+	db.exec('DROP INDEX IF EXISTS idx_db_client_conn_owner');
+};
--- a/backend/database/migrations/index.ts
+++ b/backend/database/migrations/index.ts
@@ -33,6 +33,7 @@ import * as migration031 from './031_create_db_client_tables';
 import * as migration032 from './032_seed_copilot_provider';
 import * as migration033 from './033_seed_codex_provider';
 import * as migration034 from './034_seed_qwen_provider';
+import * as migration035 from './035_add_owner_to_db_client_connections';
 
 // Export all migrations in order
 export const migrations = [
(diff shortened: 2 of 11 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks a lot for this PR — the ownership and redaction coverage is comprehensive and consistent with the admin-override pattern used in earlier PRs. I audited the diff; a few items need to be addressed before this can be merged:

## Blockers

### 1. `bun run check` fails at `backend/ws/db-client/connections.ts:183`

```
Property 'id' does not exist on type '{ ...connectionInputSchema } | { id: string }'.
```

Because `connectionTestSchema` is a `t.Union([...])`, TypeScript doesn't narrow `data` after `isInput(data)` returns false. Line 184 already uses the cast `(data as { id: string }).id`, but line 183 does not:

```ts
// current:
requireDbClientConnectionAccess(conn, data.id);
return connectionManager.test({ id: (data as { id: string }).id });

// please make it consistent, e.g.:
const id = (data as { id: string }).id;
requireDbClientConnectionAccess(conn, id);
return connectionManager.test({ id });
```

I confirmed `main` is clean (0 errors) and this branch fails with the error above, so it's a regression introduced by this PR.

### 2. SSH passphrase is silently overwritten to NULL when a user edits a connection

Because of the redaction, the frontend (`ConnectionForm.svelte`) opens the edit form with `sshPassphrase = ''`. The validator forces the user to re-enter `password` and `privateKey`, but **not** `passphrase`. On submit, in `db-client-connection-queries.ts` `update()`:

```ts
const merged = {
  ...existingValues,
  ...patch.ssh   // passphrase: '' from the now-blank form
};
push('ssh_passphrase', merged.passphrase || null);  // -> NULL, the saved passphrase is wiped
```

Suggested fix (either is fine, ideally both):
- **Frontend**: in `buildInput()`, send `passphrase: sshPassphrase || undefined` (same treatment as `password: password || undefined`).
- **Backend**: in `update()`, treat an empty string for secret fields (`ssh_password`, `ssh_private_key`, `ssh_passphrase`) as "unchanged" rather than "clear it".

### Existing connections become invisible to non-admins

Migration 035 adds `owner_user_id` with a default of `NULL`. Consequence: connections that existed before the migration are only accessible to admins (because `owner_user_id IS NULL` doesn't match any non-admin `userId`). This is reasonable since there is no owner signal in legacy data, but per our security-PR guidelines this needs to be surfaced explicitly under `## Security impact` or `## Notes` so operators know to either:

- Communicate to users that legacy connections need to be re-created, **or**
- Run a manual reassignment via SQL after upgrade.

- `markUsedForUser` is added in `db-client-connection-queries.ts` but never called — `connectionManager.get()` still uses the non-scoped `markUsed`. It's safe because it's only reached via gated paths, but the new method is effectively dead code.
- In `db-client:update` and `db-client:delete`, `requireDbClientConnectionAccess` is called in the handler and then `updateForUser` / `deleteForUser` call `ensureAccess` again. Not wrong, just a redundant double-check that could be consolidated.
- Trailing blank line at the end of `backend/ws/db-client/structure.ts` (cosmetic).

DeryFerd:
Thanks for the careful review! I pushed a follow-up commit addressing the blockers: the test handler now narrows the connection id before access checks, redacted SSH/DB secrets are preserved on edit instead of being overwritten by empty strings, and I updated the PR description with the legacy owner_user_id = NULL migration note. I also removed the unused markUsedForUser helper while I was there. Local note: git diff --check passes, but I could not run check on this machine because Bun is not installed here.

PR #226, opened 2026-05-03
fix(files): scope file operations to accessible project paths

DeryFerd:
## Summary
This PR scopes file read/write/search routes to project paths the current user can access.

## Why?
Several `files:*` WebSocket routes accepted client-provided paths directly. In multi-user mode, that could let an authenticated user read, search, modify, upload, or delete files outside projects they are allowed to access.

## Changes
- Added a reusable file path access helper for project-contained paths.
- Require project membership for `project_path` based file routes.
- Require file paths to be inside one of the current user's accessible projects.
- Filter expanded file-tree paths so they cannot escape the selected project root.
- Applied the guard to file read, browse, list, content, write, create, rename, duplicate, upload, delete, search, and replace routes.

## Validation
- Ran `git diff --check`.

### DIFF
--- a/backend/files/path-browsing.ts
+++ b/backend/files/path-browsing.ts
@@ -1,4 +1,5 @@
 import { join, extname } from 'path';
+import { existsSync } from 'fs';
 
 import { debug } from '$shared/utils/logger';
 
@@ -365,6 +366,25 @@ export async function handleWindowsDrives(): Promise<PathBrowseData> {
 			path: drivePath,
 			modified: new Date().toISOString() // Use current time for drives
 		});
+	}
+
+	if (drives.length === 0) {
+		for (let code = 65; code <= 90; code++) {
+			const drive = String.fromCharCode(code) + ':';
+			const drivePath = drive + '\\';
+			try {
+				if (existsSync(drivePath)) {
+					drives.push({
+						name: `${drive} Drive`,
+						type: 'directory',
+						path: drivePath,
+						modified: new Date().toISOString()
+					});
+				}
+			} catch {
+				// Skip unavailable drives.
+			}
+		}
 	}
 
 	// If no drives found, throw error
--- a/backend/ws/files/path-access.ts
+++ b/backend/ws/files/path-access.ts
@@ -18,11 +18,18 @@ export function requireProjectPathAccess(conn: WSConnection, projectPath: string
 	if (!project) {
 		throw new Error('Project not found');
 	}
+	if (ws.getRole(conn) === 'admin') {
+		return project;
+	}
 	return requireProjectAccess(conn, project.id);
 }
 
 export function requireFilePathAccess(conn: WSConnection, filePath: string): string {
 	const normalizedPath = resolve(filePath);
+	if (ws.getRole(conn) === 'admin') {
+		return normalizedPath;
+	}
+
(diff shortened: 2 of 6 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks for the fix — the analysis is on point and the guard helpers (`requireProjectPathAccess`, `requireFilePathAccess`, `filterAccessibleExpandedPaths`) land in the right places. Before we can merge, though, this PR introduces a regression that needs to be addressed.

## Regression

Screenshot 2026-05-08 at 21 53 05

After applying this patch, the **"Select Project Folder"** dialog (`FolderBrowser.svelte`) fails to load any directory — it shows `Error Loading Directory — File path is outside accessible projects`. This breaks the entire add-project flow, even for admin users.

**Why:** `files:browse-path` is used by the folder picker to navigate the filesystem **before** a project exists, with inputs like `'drives'`, `'home'`, `'.'`, or absolute paths the user is browsing toward. The new `requireFilePathAccess` guard at `backend/ws/files/read.ts:124` rejects all of these unconditionally because they're not inside any existing project.

**Repro:** open the "Add Project" dialog → the dialog cannot list `home`, `drives`, or any path. Screenshot attached above.

## Note on tooling

Using AI to draft patches is welcome — but please install Bun and **exercise the change in a running app** before opening the PR. Static checks don't catch UX regressions like this one, and the PR description noted you couldn't run `bun run check` locally. For security fixes especially, manual end-to-end verification is non-negotiable, otherwise a hardening patch can ship a worse usability issue than the bug it closes.

Happy to pair on the fix or take it over if it's blocking you — just let me know.

DeryFerd:
My bad, I will run and fix it now, Thanks for feedback!

Aprreciate it

DeryFerd:
Thanks for catching this, and fair feedback on the local verification gap.

I pushed a follow-up commit that restores `files:browse-path` as a pre-project folder-picker route, so `FolderBrowser` can browse `home`, `drives`, `.`, and candidate filesystem paths before project membership exists.

The file read/write/search hardening remains in place for the routes that operate on project files. I also added a Windows drive-listing fallback for environments where `wmic` returns no drives.

Verification now run locally with Bun installed:
- Exercised `handlePathBrowsing("home")` successfully on Windows.
- Exercised `handlePathBrowsing("drives")` successfully on Windows.

ArgaFairuz:
Hi @DeryFerd — thanks for the patch. I picked it up to close a few remaining gaps surfaced while testing the folder picker. Pushed one follow-up commit on top of yours:

**What the follow-up adds**
- `files:delete-directory` (empty-only) and `files:rename-directory` so the FolderBrowser can manage candidate project folders without re-introducing the broad `files:delete`/`files:rename` surface.
- Loosens `files:create-directory` via a new picker-style guard (`requireSharedFilePathAccess`) — paths outside every registered project are allowed, paths inside another user's project are rejected.
- Ancestor guard: also rejects paths that *contain* another user's project, so a member can't rename/delete the parent dir of someone else's project root.
- Admin role bypasses the project-scope guards entirely (rename/delete/create anywhere).
- Workaround for `Bun.file(dir).exists()` returning `false` on directories — the rename route uses `fs.rename` directly.
- Frontend: rename dialog + pencil button in FolderBrowser, delete switched to the empty-only route.

PR #227, opened 2026-05-03
fix(tunnel): restrict tunnel routes to admins

DeryFerd:
## Summary
This PR restricts all `tunnel:*` WebSocket routes to admin users.

## Why
Tunnel management can expose local services, store Cloudflare tunnel tokens, mutate ingress rules, and log out local cloudflared auth. Those operations are privileged and should not be available to regular authenticated users.

## Changements
- Added an admin-only route prefix for `tunnel:*`.
- Kept the existing explicit admin-only route list intact.
- Ensured current and future tunnel routes are covered without needing to enumerate every action manually.

## Validation
- Ran `git diff --check`.

### DIFF
--- a/backend/auth/permissions.ts
+++ b/backend/auth/permissions.ts
@@ -38,6 +38,8 @@ export const ADMIN_ONLY_ROUTES = new Set([
 	'system-tools:install-session'
 ]);
 
+const ADMIN_ONLY_ROUTE_PREFIXES = ['tunnel:'];
+
 /**
  * Check if a route action is allowed for the given auth state.
  * In no-auth mode, all routes are allowed (bypasses authentication check).
@@ -63,7 +65,10 @@ export function checkRouteAccess(
 	}
 
 	// Admin-only routes
-	if (ADMIN_ONLY_ROUTES.has(action) && role !== 'admin') {
+	const requiresAdmin = ADMIN_ONLY_ROUTES.has(action) ||
+		ADMIN_ONLY_ROUTE_PREFIXES.some((prefix) => action.startsWith(prefix));
+
+	if (requiresAdmin && role !== 'admin') {
 		return { allowed: false, error: 'Admin access required' };
 	}
 


ArgaFairuz:
Hi @DeryFerd, thanks for the patch. Before I can review, please install Bun and **exercise the change in a running app** end-to-end — the PR description notes you couldn't run `bun run check` locally, and PR #226 turned out to ship a UX regression for the same reason.

Using AI to draft patches is welcome, but for security-related changes especially, static checks alone don't catch regressions. Please run `bun run check` and `bun run lint`, manually verify the affected flows in the UI, and update the PR description with what you tested. I'll re-review once that's done.

DeryFerd:
Thanks for the guidance. I re-tested this PR in a running local app after installing Bun.

Validation performed:
- `bun run lint` passes.
- Started the app locally with `bun run dev`.
- Completed setup in `With Login` mode with an admin user.
- Verified the admin user can see admin-only Settings sections, including Engines, Security, Maintenance, System Tools, and Team.
- Verified the admin user can open Settings -> Tunnel without an `Admin access required` error.
- Created a member user via invite in a separate/incognito session.
- Verified the member user does not see admin-only Settings sections.
- Verified the member user can reach the Tunnel settings UI, but tunnel actions are denied by the backend with `Admin access required`.

So the backend `tunnel:*` admin guard is active, while admin access to the tunnel settings flow remains intact. I agree this needed an actual running-app check; thanks for calling it out.

Adding the member-side UI verification screenshot here as well.

This is from the invited member account. Admin-only Settings sections are hidden, and triggering the tunnel flow returns `Admin access required`, which matches the intended backend `tunnel:*` guard behavior.

image

ArgaFairuz:
Hi @DeryFerd, my apologies — I've changed my mind after a deeper look at the diff together with the surrounding Settings and Public Tunnel UI. Thanks for running the end-to-end validation I asked for earlier; that part was done correctly. Closing this PR (not asking for changes) because the **scope and pattern** need to shift, and it's cleaner to land a fresh PR than to retrofit this one.

## Why I'm closing

Two issues:

### 1. The prefix-based gate over-restricts

`ADMIN_ONLY_ROUTE_PREFIXES = ['tunnel:']` blocks every `tunnel:*` route including `tunnel:quick:start/stop` and the read endpoints. That's stricter than we actually want:

- **Quick Tunnel** is ephemeral, uses a random `*.trycloudflare.com` URL, stores no Cloudflare credentials, and persists no config. It's the "share my dev server" use case. A member already has terminal access inside their project — denying them Quick Tunnel doesn't change their threat surface, it just removes a convenience.
- **Read endpoints** (`tunnel:status`, `tunnel:remote:config:list`, `tunnel:local:config:list`, `tunnel:local:auth-status`, `tunnel:remote:ingress`) are benign — `tunnel:remote:config:list` already strips tokens in `config.ts:42`. Blocking them adds noise without raising the security bar.

The existing pattern for `engine:*` (see the comment in `permissions.ts:39-41`) explicitly states reads stay open and only mutations are admin-gated. We should stay consistent with that pattern instead of introducing a second mechanism (prefix matching).

### 2. The frontend has Tunnel surfaces beyond Settings

Even if we kept the prefix approach, there are entry points outside Settings that would break for members:

- `TunnelButton` in `DesktopNavigator.svelte:303, 361` and `MobileNavigator.svelte:161` is rendered for all users and polls `tunnel:status` via `tunnelStore`.
- `TunnelModal` mounted alongside the button calls `tunnel:remote:config:list`, `tunnel:local:config:list`, `tunnel:local:auth-status`.

That's the same class of UX regression as PR #226 — members would see error toasts from a feature they're not supposed to interact with, and we wouldn't catch it without testing as a non-admin.

## What I'm doing instead

I'll open a fresh PR that covers the full fix:

1. **Backend** — drop the prefix mechanism, enumerate the privileged tunnel routes explicitly in `ADMIN_ONLY_ROUTES` (same shape as the `engine:*` block). Quick Tunnel + reads stay open.
2. **Frontend — Settings → Tunnel** — hide the menu entry, `<TunnelSettings />` mount, and the active-section branch for non-admin users (consistent with `engines` / `system-tools` / `team`).
3. **Frontend — TunnelModal** — keep visible for all users (members still need Quick Tunnel from the navigator). Hide the "Settings → Tunnel" hint link for non-admin. Gate the stop/manage controls in `TunnelActive` for `type === 'remote' | 'local'` behind `isAdmin` so members can't trigger backend `Admin access required` errors.

Sorry again for the back-and-forth — the right pattern only became obvious to me on the second review. The security intent of your PR was correct; we just need a different shape for the fix.

PR #228, opened 2026-05-03
fix(projects): prevent implicit project membership by path

DeryFerd:
This PR prevents users from implicitly joining an existing project just by creating a project with the same filesystem path.

## Why this
Project access checks rely on membership in `user_projects`. Previously, `projects:create` would automatically add the current user to an existing project when the submitted path matched. In multi-user mode, that could allow a user who knows or guesses a project path to grant themselves access.

## Changes
- Preserve the existing behavior for users who already have access to the project.
- Reject duplicate project paths when the current user is not already a member.
- Update comments to describe the safer behavior clearly.

## Validation
- Ran `git diff --check`.

### DIFF
--- a/backend/ws/projects/crud.ts
+++ b/backend/ws/projects/crud.ts
@@ -3,7 +3,7 @@
  *
  * HTTP endpoints for project management:
  * - List all projects (per-user)
- * - Create new project (or join existing)
+ * - Create new project (or return an existing project already joined by the user)
  * - Get project by ID
  * - Update project info
  * - Delete project (remove user association, cleanup if orphaned)
@@ -34,7 +34,7 @@ export const crudHandler = createRouter()
 		return projects;
 	})
 
-	// Create new project (or join existing by path)
+	// Create new project (or return an existing project already joined by the user)
 	.http('projects:create', {
 		data: t.Object({
 			name: t.String({ minLength: 1 }),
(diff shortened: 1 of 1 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks for the patch. Before I can review, please install Bun and **exercise the change in a running app** end-to-end — the PR description notes you couldn't run `bun run check` locally, and PR #226 turned out to ship a UX regression for the same reason.

Using AI to draft patches is welcome, but for security-related changes especially, static checks alone don't catch regressions. Please run `bun run check` and `bun run lint`, manually verify the affected flows in the UI, and update the PR description with what you tested. I'll re-review once that's done.

DeryFerd:
Alright, I will fix it soon, thanks!

DeryFerd:
Thanks for the guidance. I re-tested this PR in a running local app after installing Bun.

Validation performed:
- `bun run lint` passes.
- Started the app locally with `bun run dev`.
- Completed setup in `With Login` mode with an admin user.
- Verified the admin user can add/import a local project path successfully.
- Created a member user via invite in a separate/incognito session.
- As the member user, tried to add the same filesystem path that was already registered by the admin.
- Verified the app rejects the member attempt instead of implicitly adding the member to the existing project.
- Refreshed the member session and confirmed the admin project does not appear in the member project list.

This confirms the implicit membership-by-path behavior is blocked while normal admin project import still works.

Here are the screenshots (The dark one is admin, the light is member):

Screenshots:

| Admin | Member |
| --- | --- |
| Admin project list | Member project list |

PR #235, opened 2026-05-04
refactor(engine): extract shared session-fork primitives

DeryFerd:
### Context

Three engine adapters — Copilot, Codex, and Qwen — each implement their own session-fork workaround because their respective SDKs don't yet expose a native `forkSession()` API. While the on-disk formats differ (Copilot uses a directory with YAML + JSONL, Codex uses a single JSONL in a date-tree, Qwen uses a single JSONL in a project-scoped directory), the *operational pattern* is identical across all three:

1. Check if the source exists → return `false` if missing
2. Remove destination if it already exists (clean re-fork)
3. Copy source → destination
4. Patch embedded identifiers in the copy
5. Log the fork result and return `true`

Steps 1–3 and 5 are pure file-system operations with no engine-specific logic. Yet each adapter reimplements them from scratch — `fs.cpSync`, `fs.rmSync`, `fs.mkdirSync`, `fs.writeFileSync`, etc. — leading to ~70 lines of duplicated boilerplate across the three files.

### What this PR does

Introduces [backend/engine/adapters/shared/session-fork.ts](cci:7://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/shared/session-fork.ts:0:0-0:0) — a small utility module that provides the common primitives:

| Primitive | Purpose | Used by |
|-----------|---------|---------|
| [copyDirectory(src, dst)](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/shared/session-fork.ts:16:0-29:1) | Recursive directory copy with pre-clean | Copilot |
| [copyFile(src, dst)](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/shared/session-fork.ts:31:0-47:1) | Single-file copy with mkdir + pre-clean | Codex, Qwen |
| [patchFileContent(path, old, new)](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/shared/session-fork.ts:49:0-58:1) | Blanket string replace (safe for UUIDs) | Codex, Qwen |
| [patchFileRegex(path, pattern, replacement)](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/shared/session-fork.ts:60:0-69:1) | Targeted regex replace (safe for YAML) | Copilot |
| [patchJsonlFirstLine(path, mutator, label)](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/shared/session-fork.ts:71:0-95:1) | Parse-mutate-rewrite first JSONL line | Copilot |
| [logForkResult(engine, src, fork, ok)](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/shared/session-fork.ts:97:0-108:1) | Unified debug.log / debug.warn | All three |

Each adapter now imports what it needs and keeps only its engine-specific logic (path resolution, patch strategy). The per-engine [sessionStateExists()](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/qwen/session-fork.ts:61:0-63:1) and [getSessionStatePath()](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/copilot/session-fork.ts:35:0-37:1) functions remain untouched — they're inherently engine-specific.

### Before vs After (line counts)

| Adapter | Before | After | Delta |
|---------|--------|-------|-------|
| [copilot/session-fork.ts](cci:7://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/copilot/session-fork.ts:0:0-0:0) | 112 lines | 82 lines | −30 |
| [codex/session-fork.ts](cci:7://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/codex/session-fork.ts:0:0-0:0) | 150 lines | 146 lines | −4 |
| [qwen/session-fork.ts](cci:7://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/qwen/session-fork.ts:0:0-0:0) | 98 lines | 96 lines | −2 |
| [shared/session-fork.ts](cci:7://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/shared/session-fork.ts:0:0-0:0) | — | 104 lines | +104 |

Net: +68 lines, but the shared module is reusable and testable in isolation. The real win is that future engine adapters (or changes to the copy/patch logic) only need to touch one place.

### What didn't change

- No behavioral changes — all three forks produce identical results
- The `TODO` comments about native `forkSession()` are preserved
- Engine-specific path resolution ([findRolloutFile](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/codex/session-fork.ts:48:0-84:1), [sanitizeCwd](cci:1://file:///D:/Portfolio%20Data/Learning%20PR/clopen/backend/engine/adapters/qwen/session-fork.ts:40:0-50:1), etc.) stays in each adapter
- Codex still logs the destination path in its own debug line (engine-specific detail not worth abstracting)

### Future

When all three SDKs gain native fork APIs, both this shared module *and* the per-engine session-fork files can be deleted entirely — the comment at the top of the shared module says exactly this.

### DIFF
--- a/backend/engine/adapters/codex/session-fork.ts
+++ b/backend/engine/adapters/codex/session-fork.ts
@@ -38,6 +38,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { getCodexHomeDir } from './credential';
 import { debug } from '$shared/utils/logger';
+import { copyFile, patchFileContent, logForkResult } from '../shared/session-fork';
 
 const SESSION_DIR_NAME = 'sessions';
 
@@ -130,19 +131,15 @@ export function forkCodexSessionState(sourceThreadId: string, forkThreadId: stri
 	}
 
 	const dstPath = buildForkPath(forkThreadId);
-	const dstDir = path.dirname(dstPath);
-	fs.mkdirSync(dstDir, { recursive: true });
 
-	if (fs.existsSync(dstPath)) {
-		fs.rmSync(dstPath, { force: true });
+	if (!copyFile(srcPath, dstPath)) {
+		return logForkResult('Codex', sourceThreadId, forkThreadId, false);
 	}
 
 	// UUIDs don't collide with unrelated content, so a blanket
 	// search-and-replace on the source thread id is safe and avoids
 	// parsing the (undocumented) JSONL schema line-by-line.
-	const content = fs.readFileSync(srcPath, 'utf-8');
-	const patched = content.split(sourceThreadId).join(forkThreadId);
-	fs.writeFileSync(dstPath, patched);
+	patchFileContent(dstPath, sourceThreadId, forkThreadId);
 
 	debug.log('engine', `Codex fork: ${sourceThreadId} → ${forkThreadId} (${dstPath})`);
 	return true;
--- a/backend/engine/adapters/copilot/session-fork.ts
+++ b/backend/engine/adapters/copilot/session-fork.ts
@@ -29,7 +29,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import os from 'node:os';
-import { debug } from '$shared/utils/logger';
+import { copyDirectory, patchFileRegex, patchJsonlFirstLine, logForkResult } from '../shared/session-fork';
 
 const SESSION_STATE_DIR = path.join(os.homedir(), '.copilot', 'session-state');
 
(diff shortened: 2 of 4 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks a lot for the audit — the duplication you spotted is real and the read-through across three adapters was useful. After checking against `backend/engine/README.md` §9.10 and the adapter taxonomy in §2.6, I'd like to keep the per-adapter shape rather than land this. Sharing the reasoning so it's auditable.

## Why

**1. The per-adapter shape is deliberate.** README §9.10 (~line 1394) pins the migration target:

> When an SDK eventually adds a native `forkSession`, the migration is a one-liner: delete that adapter's `session-fork.ts`, drop the fork block in `stream.ts`, and pass the SDK flag the same way Claude and OpenCode already do.

Once a `shared/` module sits between them, deleting (e.g.) `copilot/session-fork.ts` orphans `copyDirectory` / `patchFileRegex` / `patchJsonlFirstLine` in `shared/`. One deletion site becomes three.

**2. The shared surface is asymmetric.** Half the primitives are single-consumer:

| Primitive             | Consumers     |
|-----------------------|---------------|
| `copyDirectory`       | Copilot only  |
| `patchFileRegex`      | Copilot only  |
| `patchJsonlFirstLine` | Copilot only  |
| `copyFile`            | Codex, Qwen   |
| `patchFileContent`    | Codex, Qwen   |
| `logForkResult`       | All three     |

The Copilot-only helpers are just relocated away from their rationale comments (the `patchWorkspaceYaml` / `patchEventsJsonl` JSDocs explaining *why* regex-vs-parse).

**3. `adapters/shared/` isn't in the §2.6 taxonomy.** §2.6 is strict about canonical names; adding a sibling directory needs a README update first.

**4. Behavioural regression in the Qwen log.** `qwen/session-fork.ts:81` currently logs `... not found at ${srcPath}`; `logForkResult` drops the path. README §9.10 calls out `sanitizeCwd()` drift as a sharp edge — the resolved path is the first signal you want when the fork silently skips.

**5. Net +68 lines** justified as "reusable for future adapters", but there are five engines today and the three on-disk workarounds are slated for deletion the moment each SDK ships native fork.

## Final note — the three workarounds aren't actually the same operation

Surface-wise they look identical (copy → patch → log), but each adapter solves a different problem against a different on-disk layout:

- **Copilot** — a directory with YAML + JSONL; needs two targeted in-place rewrites (`workspace.yaml#id` regex, `events.jsonl` first-line `data.sessionId`) to preserve formatting.
- **Codex** — single JSONL inside a `YYYY/MM/DD` date-tree keyed by filename suffix; needs date-tree walk + blanket UUID replace.
- **Qwen** — single JSONL inside a `sanitizeCwd()`-derived project dir; needs cwd as input and blanket replace.

The truly shared logic is roughly "copy a thing, then mutate ids" — collapsing it to common primitives forces the engine-specific rationale (why regex, why blanket replace, why date-tree, why `sanitizeCwd`) to move away from where it's read. Keeping each adapter self-contained is what makes the §9.10 sharp-edges actually readable next to the code they protect.

Thanks again for digging into this area.

PR #241, opened 2026-05-12
fix(security): standardize access denial error messages

DeryFerd:
Access denial errors previously distinguished "doesn't exist" from "no access," giving attackers a probe oracle to enumerate resources by observing which error variant they got back.

Standardized all access denial paths to throw or emit `"Access denied"` so the message text no longer discloses resource existence.

## Changes

22 sites across 10 files:

- `backend/auth/auth-service.ts` (2 locations) — `assignProjectToUser`, `unassignProjectFromUser`
- `backend/ws/access.ts` (5 locations) — `requireProjectAccess`, `requireSessionAccess`, `requireMessageAccess`
- `backend/ws/files/path-access.ts` (4 locations) — `requireProjectPathAccess`, `requireFilePathAccess`, `requireSharedFilePathAccess`
- `backend/ws/files/read.ts` (1 location)
- `backend/ws/files/watch.ts` (1 location)
- `backend/ws/sessions/crud.ts` (1 location)
- `backend/ws/terminal/access.ts` (3 locations) — `requirePtySessionAccess`, `requireTerminalStreamAccess`, `requireTerminalLookupAccess`
- `backend/ws/terminal/session.ts` (2 locations)
- `backend/ws/terminal/persistence.ts` (1 location) — `terminal:reconnect` emit
- `backend/ws/preview/browser/console.ts` (2 locations)

## Security impact

**Threat:** any authenticated member or unauthenticated visitor with WS access could distinguish "resource doesn't exist" from "resource exists but you have no access" by reading the error message text. This enabled enumeration of project IDs and paths, session IDs, message IDs, terminal session/stream IDs, and file paths inside other users' projects.

**Closed by this PR:** every access denial path now returns `"Access denied"` uniformly across throws and event emits. The message text no longer discloses resource existence.

**Out of scope:** timing-based existence inference (the cryptographic token storage already defeats this — see review discussion); HTTP status code variation (responses wrap uniformly as \`{ success: false }\` regardless).

## Testing

- \`bun run check\` passes (0 errors, 0 warnings).
- \`bun run lint\` passes clean.
- Verified the frontend dead-terminal-tab auto-close (\`frontend/stores/features/terminal.svelte.ts:498\`) is unaffected — it keys on \`"Session not found or PTY not available"\` emitted from \`backend/ws/terminal/stream.ts:40\`, which signals a runtime PTY write failure rather than an authorization decision and is intentionally left unchanged.

Addresses FINDING-004 from repository audit.

### DIFF
--- a/backend/auth/auth-service.ts
+++ b/backend/auth/auth-service.ts
@@ -492,11 +492,11 @@ export function listUserProjects(userId: string): Project[] {
 export function assignProjectToUser(userId: string, projectId: string): boolean {
 	const user = authQueries.getUserById(userId);
 	if (!user) {
-		throw new Error('User not found');
+		throw new Error('Access denied');
 	}
 	const project = projectQueries.getById(projectId);
 	if (!project) {
-		throw new Error('Project not found');
+		throw new Error('Access denied');
 	}
 	if (projectQueries.userHasProject(userId, projectId)) {
 		return false;
@@ -513,11 +513,11 @@ export function assignProjectToUser(userId: string, projectId: string): boolean 
 export function unassignProjectFromUser(userId: string, projectId: string): boolean {
 	const user = authQueries.getUserById(userId);
 	if (!user) {
-		throw new Error('User not found');
+		throw new Error('Access denied');
 	}
 	const project = projectQueries.getById(projectId);
 	if (!project) {
-		throw new Error('Project not found');
+		throw new Error('Access denied');
 	}
 	if (!projectQueries.userHasProject(userId, projectId)) {
 		return false;
--- a/backend/ws/access.ts
+++ b/backend/ws/access.ts
@@ -7,12 +7,12 @@ export function requireProjectAccess(conn: WSConnection, projectId: string): Pro
 	const userId = ws.getUserId(conn);
 	const hasAccess = projectQueries.userHasProject(userId, projectId);
 	if (!hasAccess) {
-		throw new Error('Project not found');
+		throw new Error('Access denied');
 	}
 
 	const project = projectQueries.getById(projectId);
 	if (!project) {
-		throw new Error('Project not found');
+		throw new Error('Access denied');
 	}
 
 	return project;
@@ -32,13 +32,13 @@ export function requireCurrentProjectAccess(conn: WSConnection): {
 export function requireSessionAccess(conn: WSConnection, sessionId: string): ChatSession {
 	const session = sessionQueries.getById(sessionId);
 	if (!session) {
-		throw new Error('Session not found');
+		throw new Error('Access denied');
 	}
 
 	const userId = ws.getUserId(conn);
 	const hasAccess = projectQueries.userHasProject(userId, session.project_id);
 	if (!hasAccess) {
-		throw new Error('Session not found');
+		throw new Error('Access denied');
 	}
 
 	return session;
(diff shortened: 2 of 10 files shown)


ArgaFairuz:
Hi @DeryFerd, thank you for this contribution — really appreciate the careful audit of the auth path. I have a few thoughts I'd like to share, and I'd genuinely value your perspective before we settle on the final shape.

## Standardized error messages — looks good

The direction is right: distinguishing "doesn't exist" from "no access" is itself a probe oracle. To make this complete, could you also cover the sites that still leak the same way?

- `backend/auth/auth-service.ts:499,520` — `"Project not found"`
- `backend/ws/files/read.ts:188` — `"Project not found"`
- `backend/ws/files/watch.ts:28` — `"Project not found"`
- `backend/ws/sessions/crud.ts:401` — `"Session not found"`
- `backend/ws/terminal/access.ts:20,52` — `"Session not found"`
- `backend/ws/terminal/session.ts:61,316` — `"Session not found"`
- `backend/ws/preview/browser/console.ts:36,72` — `"Session not found"`

## Timing-safe token comparison — would value your reasoning

I want to be upfront that I have some concerns here, but I'm genuinely open if there's a stronger argument — I might be missing something.

My main concern: `crypto.timingSafeEqual` defeats timing attacks where an attacker can iteratively steer their input toward a target byte-by-byte (the classic case is comparing a plaintext API key or an HMAC). Here, what we compare are SHA-256 hashes of the token, not the tokens themselves. Even if hash-comparison timing leaked perfectly, an attacker would only learn `sha256(valid_token)` — and recovering the original token from that requires a preimage attack on SHA-256, which is computationally infeasible.

The trade-off that worries me: the new code moves `getSessionByTokenHash` from an O(log n) index lookup to an O(n) full-table scan, and that path runs on every WS auth check. `getInviteByTokenHash` has the same shape and is reachable from the unauthenticated signup endpoint.

I also noticed `deleteSessionByTokenHash` (`auth-queries.ts:138`) still uses `WHERE` directly — if the timing concern applies to lookups, it would apply there too, so I want to make sure we have a consistent picture before adopting it as a pattern.

**I could be wrong about the threat model here.** If you have a concrete scenario where the hash-comparison timing is actually exploitable, I'd really like to hear it — if the case is strong, we keep both fixes in this PR. Otherwise, would you be open to splitting this part into a separate PR so we can land the error-message fix first while we keep discussing this one?

## Next step

Please share your thoughts by **2026-05-20**. If we don't hear back by then I'll close this PR as auto-stale — you're welcome to reopen or resubmit whenever it's convenient.

Thanks again for the careful work here.

DeryFerd:
Thanks for the detailed review - you're right about the timing-safe comparison. I went back and forth on this one, but your point about SHA-256 preimage resistance is solid. We're comparing hashes, not plaintext, so the timing attack surface is basically theoretical. And the O(n) table scan on every WebSocket auth check is a real cost for a threat that's not actually exploitable.

I've split the PR. Removed the timing-safe changes entirely and completed the error message standardization across all the locations you mentioned:

- backend/auth/auth-service.ts (lines 499, 520)
- backend/ws/files/read.ts (line 188)
- backend/ws/files/watch.ts (line 28)
- backend/ws/sessions/crud.ts (line 401)
- backend/ws/terminal/access.ts (lines 20, 52)
- backend/ws/terminal/session.ts (lines 61, 316)
- backend/ws/preview/browser/console.ts (lines 36, 72)

Everything now returns "Access denied" regardless of why access was denied. The information disclosure issue is fixed without the performance hit.

bun run check still passes. Ready for another look when you have time.

PR #249, opened 2026-05-15
fix(files): enforce 50MB size limit on write and upload operations

DeryFerd:
## What this addresses

I was looking at the file write and upload paths and noticed there's no size check anywhere. `files:write-file` takes a `content` string and writes it directly via `Bun.write()`. `files:upload-file` takes a `file.data` Uint8Array and does the same. Neither endpoint validates how much data is coming in before committing it to disk.

This means any authenticated user can send a 2GB string or upload a 500MB file and the server will just write it. Do that a few times and the disk fills up. Once the disk is full, SQLite can't write, sessions can't be saved, the whole workspace goes sideways. It's a pretty straightforward DoS and it doesn't require any special privileges — just a valid session token.

## What changed

Three files:

**`backend/files/file-size-limit.ts`** (new)
- Exports `validateFileSize(size: number)` — throws if size is negative or exceeds `MAX_FILE_SIZE`
- Exports `MAX_FILE_SIZE` constant set to 50MB (50 * 1024 * 1024 bytes)
- That's it. Single-purpose module, no dependencies.

**`backend/files/file-operations.ts`**
- Added `validateFileSize(contentSize)` call at the top of `writeFileOperation()`, using `Buffer.byteLength(content, 'utf8')` to get the actual byte size of the string
- Added `validateFileSize(file.size)` call at the top of `uploadFileOperation()`, using the size field from the upload payload
- Both checks run *before* any disk I/O, so we reject early without touching the filesystem

**`backend/files/file-size-limit.test.ts`** (new)
- 5 tests: accepts files at the limit, rejects files over the limit, accepts small/zero-byte files, rejects very large files, rejects negative sizes

## Why 50MB

Picked 50MB as a reasonable ceiling for typical workspace files. Source code, config files, small assets — none of these should be anywhere near 50MB. If someone is legitimately editing a 100MB JSON file or uploading a large binary, the limit can be raised or made configurable later. The constant is in one place so it's a single-line change.

I went with a hard limit rather than a configurable one for this first pass because adding config UI, validation, and migration for a configurable limit is a bigger change. Happy to follow up with a configurable version if the maintainers think it's needed.

## Tests

```
bun test backend/files/file-size-limit.test.ts

✓ accepts files within the default size limit
✓ rejects files exceeding the default size limit
✓ accepts small files
✓ rejects very large files
✓ rejects negative sizes

5 pass, 0 fail
```

Lint passes clean. No changes to existing behavior for files under the limit.

## Things I considered but didn't do

- **Per-user or per-project quotas** — overkill for this PR. The goal is to stop disk exhaustion, not implement fair usage policies. That's a separate feature.
- **Streaming writes with mid-write abort** — Bun.write() doesn't support streaming anyway, and the content is already fully in memory by the time it reaches the write function. The size check happens before the write, so there's no partial write to clean up.
- **Different limits for write vs upload** — kept it simple with one constant. If upload needs a different limit than write, we can add a second parameter to `validateFileSize()` later.
- **Error message localization** — the error message is plain English ("File size exceeds maximum allowed size of 50MB"). This matches how other errors in `file-operations.ts` are formatted.

### DIFF
--- a/backend/files/file-operations.ts
+++ b/backend/files/file-operations.ts
@@ -148,6 +148,8 @@ export async function createFileOperation(filePath: string, content: string = ''
 		throw new Error('File path is required');
 	}
 
+	validateFileSize(Buffer.byteLength(content, 'utf8'));
+
 	try {
 		// Normalize path for Windows only
 		const normalizedFilePath = process.platform === 'win32' ?
@@ -372,8 +374,9 @@ export async function uploadFileOperation(file: { name: string; type: string; si
 		throw new Error('Target path is required');
 	}
 
-	// Validate file size before writing
-	validateFileSize(file.size);
+	// Validate the actual byte length of the payload, not the client-supplied
+	// `file.size` field — a mismatched value would otherwise bypass the limit.
+	validateFileSize(file.data.byteLength);
 
 	try {
 		// Normalize target path for Windows only
--- a/backend/files/file-operations.ts
+++ b/backend/files/file-operations.ts
@@ -2,6 +2,7 @@ import { dirname, extname, join } from 'path';
 import { stat as fsStat } from 'node:fs/promises';
 
 import { debug } from '$shared/utils/logger';
+import { validateFileSize } from './file-size-limit';
 
 // Import Bun native functions and create wrappers for better API compatibility
 const { spawn } = Bun;
@@ -119,6 +120,10 @@ export async function writeFileOperation(filePath: string, content: string) {
 	if (typeof content !== 'string') {
 		throw new Error('Content must be a string');
 	}
+
+	// Validate content size before writing
+	const contentSize = Buffer.byteLength(content, 'utf8');
+	validateFileSize(contentSize);
 
 	try {
 		debug.log('file', 'Writing file:', { filePath, contentLength: content.length });
@@ -366,6 +371,9 @@ export async function uploadFileOperation(file: { name: string; type: string; si
 	if (!targetPath) {
 		throw new Error('Target path is required');
 	}
+
+	// Validate file size before writing
+	validateFileSize(file.size);
 
 	try {
 		// Normalize target path for Windows only
--- a/backend/files/file-size-limit.test.ts
+++ b/backend/files/file-size-limit.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, test } from 'bun:test';
+import { validateFileSize } from './file-size-limit';
+
+describe('validateFileSize', () => {
+	test('accepts files within the default size limit', () => {
+		// Default limit is 50MB
+		const withinLimit = 50 * 1024 * 1024; // 50MB
+		expect(() => validateFileSize(withinLimit)).not.toThrow();
+	});
+
+	test('rejects files exceeding the default size limit', () => {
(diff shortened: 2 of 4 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks for catching this — the validator module + early-exit shape is the right call. While auditing I found a critical bypass and two adjacent sites that need the same protection, so I pushed three follow-up commits on your branch.

## Summary
Validates the actual byte length on every user-driven write path: `upload`, `create-file`, and `replace-in-files`.

## Why
The original patch had a bypass and two coverage gaps that all reach the same `Bun.write` sink:

1. **Client-controlled size (critical)** — `uploadFileOperation` validated `file.size` from the payload (`backend/files/file-operations.ts:376`), but the bytes actually written are `file.data` (line 397). The WS schema (`backend/ws/files/write.ts:124-133`) accepts both fields independently with no cross-check, so a payload like `{ size: 1, data: <500MB Uint8Array> }` passed the limit and still wrote 500MB to disk.
2. **`createFileOperation` uncovered** — `files:create-file` (`backend/ws/files/write.ts:52-66`) accepts an optional `content` string and writes it via `Bun.write` (`backend/files/file-operations.ts:166`). Same DoS shape, different endpoint — attacker can just call `create-file` instead of `write-file`.
3. **`files:replace-in-files` uncovered** — `searchAndReplace` (`backend/ws/files/search.ts:319`) writes `content.replace(pattern, replaceWith)`. A short pattern with a long replacement (`a` → `<100KB string>`) inflates the file past the limit with no check.

## Changes
- `backend/files/file-operations.ts:377` — validate `file.data.byteLength` instead of `file.size`.
- `backend/files/file-operations.ts:151` — validate `Buffer.byteLength(content, 'utf8')` in `createFileOperation` before any I/O.
- `backend/ws/files/search.ts:320` — validate the post-expansion `newContent` byte length before `Bun.write` in `replaceInFiles`.

## Security impact
**Threat model.** Authenticated user (lowest-privilege role with workspace access). Before: could write/upload arbitrary sizes via `files:write-file`, `files:upload-file` (size-field bypass), `files:create-file`, or `files:replace-in-files` — leading to disk exhaustion, which cascades into SQLite write failures and session loss. After: all four paths reject payloads >50MB before `Bun.write` is called.

## Notes
- `bun run check`, `bun run lint`, and `bun test backend/files/file-size-limit.test.ts` all pass clean.
- 2 files changed, +9/-2 on top of the original commit.
- Existing validator unit tests still cover the function itself; the new call sites are exercised through their handlers — happy to add handler-level integration tests if you'd like.

## Follow-ups (not in this PR)
- Transport-level body limit at the Elysia/WS boundary: validation currently runs *after* the full `Uint8Array` is buffered into memory, so concurrent oversized uploads can still pressure RAM even though disk is safe. A `maxBodyLength` (or similar) would shed the bytes at parse time.
- Consider consolidating the three size constants (`backend/files/file-size-limit.ts` = 50MB, `backend/snapshot/snapshot-service.ts:36` = 5MB, `frontend/components/chat/input/composables/use-file-handling.svelte.ts:33` = 10MB) — different contexts may justify different values, but they're worth being explicit about in one shared module.

PR #250, opened 2026-05-15
fix(auth): invalidate user sessions when project access is revoked

DeryFerd:
## Problem

When an admin removes a user from a project, the database assignment gets deleted — but the user's active WebSocket sessions stay alive. They keep their in-memory auth state (userId, role, sessionTokenHash) and can keep working in that project until they naturally disconnect or their session expires, which by default is 30 days.

So if you revoke someone's access today because they left the team, they can still read files, run terminal commands, and see chat history for the next month unless they happen to close their browser. The access check only runs on reconnection and on certain operations, but the WebSocket connection itself stays authenticated.

I noticed this while looking at how `unassignProjectFromUser()` works. It removes the project assignment from the database and logs it, but nothing touches the active sessions. There's already a pattern for this — `logoutAllSessions()` and `clearAllAuth()` exist for when auth mode changes — but there's no targeted version for a single user.

## What changed

**`backend/auth/session-invalidation.ts`** (new)
- `invalidateUserSessions(userId)` — looks up all active WebSocket connections for a user, clears their auth state, and emits a `auth:force-logout-user` event so the frontend knows to kick them back to the login screen
- `getActiveSessionCountForUser(userId)` — utility, mostly for logging

**`backend/utils/ws.ts`**
- `getConnectionsForUser(userId)` — returns all active (readyState === 1) WebSocket connections for a given user. Uses the existing `userConnections` room Map that's already maintained for user-scoped event emission
- `clearAuthForConnections(connections[])` — targeted version of `clearAllAuth()`. Clears authenticated, role, and sessionTokenHash on a specific set of connections instead of everyone

**`backend/auth/auth-service.ts`**
- `unassignProjectFromUser()` now calls `invalidateUserSessions(userId)` after removing the project assignment. Wrapped in a `require()` call to avoid circular dependency with the ws module

**`backend/ws/auth/index.ts`**
- Declares the new `auth:force-logout-user` event with `userId` and `reason` fields

**`frontend/stores/features/auth.svelte.ts`**
- Added listener for `auth:force-logout-user`. Only acts if the payload userId matches the current user — other connected users ignore it. Clears currentUser, sessionToken, localStorage, and sets authState back to 'login'

**`backend/auth/session-invalidation.test.ts`** (new)
- 4 tests covering: clears auth on all connections, returns 0 when no connections, returns correct count, returns 0 for nonexistent user

## How the flow works

1. Admin removes user from project via UI
2. Backend calls `unassignProjectFromUser(userId, projectId)`
3. Database assignment is deleted
4. `invalidateUserSessions(userId)` runs:
   a. Finds all active WebSocket connections for that user
   b. Clears auth state on each connection (authenticated = false, role = null, sessionTokenHash = null)
   c. Emits `auth:force-logout-user` globally with the userId
5. Frontend receives the event, checks if it's for the current user
6. If yes: clears auth state, redirects to login
7. If no: ignores it

## Tests

```
bun test backend/auth/session-invalidation.test.ts

✓ clears auth on all connections for a user
✓ returns 0 when user has no active connections
✓ returns count of active connections for user
✓ returns 0 when user has no connections

4 pass, 0 fail
```

Lint passes clean.

## Edge cases I thought about

**User has no active connections at the time of removal.** The invalidation returns 0 and the event still fires. When the user reconnects later, their session token is still valid in the database, but the project access check will fail because the assignment was removed. They'll get an access denied error on project-specific operations rather than a clean logout. The access is still blocked, just with a slightly different error path. Not ideal but acceptable — the important thing is they can't access the project anymore.

**User is connected from multiple devices/browsers.** All connections get invalidated. `getConnectionsForUser()` returns every active connection for that userId, and `clearAuthForConnections()` clears all of them. The global event fires once and each frontend instance checks if it's the targeted user.

**Admin removes themselves from a project.** Works the same way. Their sessions get invalidated and they get logged out. They'd need to log back in and the project won't be in their list anymore.

**The `require()` call for circular dependency.** `auth-service.ts` can't do a top-level import of `session-invalidation.ts` because `session-invalidation.ts` imports from `$backend/utils/ws`, and the ws module eventually imports from auth. Using `require()` inside the function body breaks the cycle. Not pretty but it works and is a common pattern in this codebase for this specific dependency shape.

### DIFF
--- a/backend/auth/auth-service.ts
+++ b/backend/auth/auth-service.ts
@@ -523,7 +523,12 @@ export function unassignProjectFromUser(userId: string, projectId: string): bool
 		return false;
 	}
 	projectQueries.removeUserProject(userId, projectId);
-	debug.log('auth', `Project ${projectId} unassigned from user ${userId}`);
+
+	// Invalidate all active sessions for this user to enforce the access change
+	const { invalidateUserSessions } = require('./session-invalidation');
+	const clearedCount = invalidateUserSessions(userId);
+	debug.log('auth', `Project ${projectId} unassigned from user ${userId}, ${clearedCount} session(s) invalidated`);
+
 	return true;
 }
 
--- a/backend/auth/auth-service.ts
+++ b/backend/auth/auth-service.ts
@@ -523,11 +523,7 @@ export function unassignProjectFromUser(userId: string, projectId: string): bool
 		return false;
 	}
 	projectQueries.removeUserProject(userId, projectId);
-
-	// Invalidate all active sessions for this user to enforce the access change
-	const { invalidateUserSessions } = require('./session-invalidation');
-	const clearedCount = invalidateUserSessions(userId);
-	debug.log('auth', `Project ${projectId} unassigned from user ${userId}, ${clearedCount} session(s) invalidated`);
+	debug.log('auth', `Project ${projectId} unassigned from user ${userId}`);
 
 	return true;
 }
--- a/backend/auth/session-invalidation.test.ts
+++ b/backend/auth/session-invalidation.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, test, mock, beforeEach } from 'bun:test';
+import { invalidateUserSessions, getActiveSessionCountForUser } from './session-invalidation';
+
+// Mock the ws module
+const mockGetConnectionsForUser = mock(() => [] as any[]);
+const mockClearAuthForConnections = mock(() => 0);
+const mockEmitGlobal = mock(() => {});
+
+mock.module('$backend/utils/ws', () => ({
+	ws: {
+		getConnectionsForUser: mockGetConnectionsForUser,
+		clearAuthForConnections: mockClearAuthForConnections,
+		emit: {
+			global: mockEmitGlobal
+		}
+	}
+}));
+
+describe('invalidateUserSessions', () => {
+	beforeEach(() => {
+		mockGetConnectionsForUser.mockClear();
+		mockClearAuthForConnections.mockClear();
+		mockEmitGlobal.mockClear();
+	});
+
+	test('clears auth on all connections for a user', () => {
+		mockGetConnectionsForUser.mockReturnValue(['conn1', 'conn2', 'conn3']);
+		mockClearAuthForConnections.mockReturnValue(3);
+
+		const cleared = invalidateUserSessions('user-123');
+
+		expect(cleared).toBe(3);
+		expect(mockGetConnectionsForUser).toHaveBeenCalledWith('user-123');
+		expect(mockClearAuthForConnections).toHaveBeenCalledWith(['conn1', 'conn2', 'conn3']);
+	});
+
+	test('returns 0 when user has no active connections', () => {
+		mockGetConnectionsForUser.mockReturnValue([]);
+		mockClearAuthForConnections.mockReturnValue(0);
+
+		const cleared = invalidateUserSessions('user-456');
+
+		expect(cleared).toBe(0);
+	});
+});
+
+describe('getActiveSessionCountForUser', () => {
+	beforeEach(() => {
+		mockGetConnectionsForUser.mockClear();
+	});
+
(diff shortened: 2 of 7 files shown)


ArgaFairuz:
Thanks @DeryFerd — the "Edge cases I thought about" section made this audit much faster than it normally would be, and the threat model (revoked access persisting in-memory until natural disconnect) is real. The overall shape is right. Two things to talk through before merging, plus one adjacent surface to flag for follow-up.

At `session-invalidation.ts:29` you're emitting globally and filtering on the frontend at `auth.svelte.ts:46-57`. There's already `ws.emit.user(userId, event, payload)` at `backend/utils/ws.ts:731` that targets just that user's connections — switching to it keeps the payload narrow and removes the need for the frontend userId filter. Is there a reason the global emit is intentional? If not, I'd lean toward the targeted emitter.

The `require('./session-invalidation')` at `auth-service.ts:527` is the part I'd most like to revisit. Nothing else in `backend/auth/` uses dynamic `require()`, and the circular dep dissolves if `invalidateUserSessions(userId)` is called from the route handler at `backend/ws/auth/users.ts:82-91` instead — it already imports `ws` and emits `auth:user-projects-changed` right next door, so the invalidation belongs in the same spot. Auth-service stays a pure DB layer and the import graph stays static. I'd reconsider if there's a non-WS caller of `unassignProjectFromUser` that also needs invalidation, but I'm not seeing one.

Separately for a follow-up: `removeUser()` at `auth-service.ts:338` and `regeneratePAT()` at `auth-service.ts:428` share the exact threat model — both revoke access but leave the in-memory WS auth intact. Outside this PR's scope, but worth wiring into the same `invalidateUserSessions` once it lands.

Happy to take it on, or yours if you want — your call.

Could you take a look by May 25, 2026? If you can't respond by then, I'll close this PR as auto-stale — you can reopen anytime once you're back.

DeryFerd:
Thanks for the detailed review. I pushed a follow-up commit that applies the routing changes you suggested.

What changed:
- Switched session invalidation emit from `ws.emit.global(...)` to `ws.emit.user(userId, ...)` in `backend/auth/session-invalidation.ts`.
- Moved invalidation call out of `auth-service` and into the WS route handler (`backend/ws/auth/users.ts`) after successful `auth:unassign-project`.
- Removed the dynamic `require('./session-invalidation')` from `auth-service` so that service stays DB-focused and the import graph remains static.
- Narrowed `auth:force-logout-user` event payload to `{ reason }` and simplified frontend handling in `frontend/stores/features/auth.svelte.ts` (no user-id filtering needed on targeted emits).

Validation run:
- `bun test backend/auth/session-invalidation.test.ts`
- `bunx eslint backend/auth/session-invalidation.ts backend/auth/session-invalidation.test.ts backend/auth/auth-service.ts backend/ws/auth/users.ts backend/ws/auth/index.ts`

ArgaFairuz:
Thanks @DeryFerd — routing changes look clean. The targeted `ws.emit.user(userId, ...)` at `backend/auth/session-invalidation.ts:29` keeps the payload narrow, and the move to `backend/ws/auth/users.ts:86` keeps `auth-service` as a pure DB layer with a static import graph. Order of operations in the handler is right: clear in-memory auth state first, then emit `auth:user-projects-changed` — so any in-flight WS message hits the auth gate with the new state.

Verified `unassignProjectFromUser` has no other call sites (only `backend/ws/auth/users.ts:84`), so moving the invalidation out of the service layer doesn't leave any path uncovered. Adding `getProjectChatSessions` to the test mock in the follow-up commit is a fine workaround for Bun's session-wide `mock.module` behavior — not pretty but practical.

`bun test backend/auth/session-invalidation.test.ts` → 4/4 pass, `bun run check` and `bun run lint` clean locally. Merging.

For follow-up (separate PR, no rush): `removeUser()` at `auth-service.ts:338` and `regeneratePAT()` at `auth-service.ts:428` share the same threat model and would benefit from the same `invalidateUserSessions` wiring. Happy to take that on next.

PR #251, opened 2026-05-15
fix(auth): add in-memory mutex to prevent setup race condition

DeryFerd:
## The problem

The initial setup flow has a classic race condition. Both `auth:setup` and `auth:setup-no-auth` work the same way: first they call `needsSetup()` to check if any users exist, then they create the first admin account.

If two setup requests arrive at the same time — maybe a user double-clicked the button, or a load balancer retried the request — both can pass the `needsSetup()` check before either one creates the user. Result: two admin accounts get created instead of one.

It's a time-of-check-time-of-use (TOCTOU) gap. The check and the action aren't atomic.

## The fix

Added a simple in-memory mutex (`setup-lock.ts`) that serializes setup requests. Only one setup can run at a time.

**`backend/auth/setup-lock.ts`** (new)
- `acquireSetupLock()` — returns `true` if the lock was free and is now held, `false` if someone else is already setting up
- `releaseSetupLock()` — frees the lock. Safe to call even if it wasn't held (no-op)
- `isSetupLocked()` — check current state, mostly for debugging

**`backend/auth/auth-service.ts`**
- `createAdmin()` now acquires the lock at the top, checks `needsSetup()` inside the lock, and releases in a `finally` block. If the lock is already held, it throws "Setup already in progress" immediately.

**`backend/ws/auth/login.ts`**
- The `auth:setup-no-auth` handler does the same — acquires lock before checking `needsSetup()`, releases in `finally`.

So the sequence is now:
1. Request A arrives, acquires lock
2. Request B arrives, lock is held → throws immediately
3. Request A checks `needsSetup()` (safe, no race), creates admin
4. Request A releases lock

## Tests

```
bun test backend/auth/setup-lock.test.ts

✓ acquireSetupLock returns true when lock is available
✓ acquireSetupLock returns false when lock is already held
✓ isSetupLocked returns true after lock is acquired
✓ isSetupLocked returns false when lock is not held
✓ releaseSetupLock allows re-acquiring
✓ releaseSetupLock when not locked does not throw

6 pass, 0 fail
```

Lint is clean.

## Why an in-memory lock

Clopen is designed as a single-instance local tool. It runs on your machine, one process. There's no cluster, no load balancer, no multi-instance deployment. An in-memory boolean flag is the simplest thing that works.

If Clopen ever gets deployed in a multi-instance setup (which isn't on the roadmap), this would need to become a distributed lock — something like a SQLite advisory lock or a Redis SET NX. But for now, that's overengineering a problem that doesn't exist.

## What I didn't change

I didn't add a database-level unique constraint on the admin role or anything like that. The `needsSetup()` check is the right gate — the lock just makes it atomic. Adding DB constraints would be a belt-and-suspenders approach, but it would also require migration logic and error handling for constraint violations, which is more surface area for a scenario that's already covered by the lock.

### DIFF
--- a/backend/auth/auth-service.ts
+++ b/backend/auth/auth-service.ts
@@ -9,6 +9,7 @@ import type { Project } from '$shared/types/database/schema';
 import { generateSessionToken, generatePAT, generateInviteToken, hashToken, getTokenType } from './tokens';
 import { generateColorFromString, getInitials } from '$backend/utils/user-helpers';
 import { debug } from '$shared/utils/logger';
+import { acquireSetupLock, releaseSetupLock } from './setup-lock';
 
 /** Default session lifetime in days */
 const DEFAULT_SESSION_DAYS = 30;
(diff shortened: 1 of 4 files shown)


ArgaFairuz:
Thanks @DeryFerd — I appreciate the instinct to dig into setup-flow race conditions, and the threat model you wrote up in the description made the audit straightforward. There's one fundamental thing I'd like to resolve before we merge, though.

I don't think the race condition this PR closes can actually happen in the current runtime. `bun:sqlite` calls are synchronous — `db.prepare(...).run()` and `.get()` in `backend/database/queries/auth-queries.ts:38-76` block the event loop rather than returning a promise. In both `auth:setup` (`backend/ws/auth/login.ts:42-58`) and `auth:setup-no-auth` (`backend/ws/auth/login.ts:68-89`), there's no `await` between `needsSetup()` and the user-creation path — `createAdmin()` in `auth-service.ts:156-191` is fully synchronous end-to-end. Single-threaded JS event loop plus zero await points means the check-and-write is already atomic by construction: a second request can only start running after the first has finished its synchronous block, and by then `needsSetup()` returns `false`.

The double-click scenario in the description reads as two sequential submits, not parallel ones — the second handler invocation won't run until the first has fully returned, and it'll find the user already exists.

What would flip my position: a concrete code path where the check and the write are actually separated by an `await`, or a threat model where the single-process assumption breaks (a planned migration to multi-instance, a worker pool, etc.). If either applies, I'd love to talk through the right shape — my hunch is that defense-in-depth here belongs at the DB layer (e.g. a partial unique index on `role = 'admin'` in `migrations/024_create_users_table.ts:10-20`) rather than in-memory, since a module-level boolean won't survive a future multi-instance deployment anyway.

One more thing on the tests: `setup-lock.test.ts` covers the lock primitive in isolation, but there's no concurrent-setup test that demonstrates the race condition existed or that it's now closed. Per CONTRIBUTING.md's tests guidance, a security-shaped fix wants a case that fails before the patch and passes after — right now the new tests would pass even if the lock had no callers.

Could you take a look by May 25, 2026? If you can't respond by then, I'll close this PR as auto-stale per CONTRIBUTING.md — you can reopen anytime once you're back.

DeryFerd:
Thanks for the careful breakdown. I re-validated the flow and I agree with your assessment: in the current Bun + sync-SQLite path there is no await boundary between setup check and admin create, so this in-memory mutex is not providing a real protection in today’s runtime model.\n\nI’m closing this PR to avoid carrying the wrong abstraction. If we do future hardening for multi-instance deployment, I’ll open a separate PR with a DB-level invariant (e.g. partial unique index for admin role) and a corresponding integration test.

PR #252, opened 2026-05-15
fix(memory): add centralized cleanup registry for project state Maps

DeryFerd:
## Problem

There are a few places in the frontend where we keep in-memory state keyed by project ID or path, and we never clean it up when a project is deleted.

- `FilesPanel.svelte` keeps `projectFileStates` — a Map of panel layout, open tabs, and scroll positions per project.
- `PreviewPanel.svelte` keeps `projectPreviewState` — browser preview state (URL, device size, rotation) per project.
- `global-stream-monitor.ts` keeps `notifiedToolUseIds` — a Set of notification IDs to prevent duplicates.

These Maps are at module level so they survive component destruction (which is intentional — it preserves state when you switch between mobile and desktop layout). But when a project is removed from the workspace, its entries stay in these Maps forever. The only way to reclaim that memory is to refresh the page.

For someone who creates and deletes projects regularly, this adds up. It's not going to crash the browser tomorrow, but it's the kind of slow creep that makes the app feel sluggish after a few weeks of use.

## How I fixed it

Instead of having each component try to clean up after itself (which is fragile and easy to forget), I added a centralized registry pattern.

**`frontend/utils/project-state-cleanup.ts`** (new)
- `registerProjectCleanup(fn)` — modules call this to register their cleanup function
- `cleanupProjectState(projectId, projectPath)` — executes all registered cleanup functions. Called when a project is removed.

**Each module registers its own cleanup:**
- `FilesPanel.svelte` — deletes the project's entry from `projectFileStates` using the project path
- `PreviewPanel.svelte` — deletes the project's entry from `projectPreviewState` using the project ID
- `global-stream-monitor.ts` — clears `notifiedToolUseIds`. Since these IDs are global and not keyed by project, a full clear is the safest option. Worst case is a duplicate notification once, which is better than a memory leak.

**`frontend/stores/core/projects.svelte.ts`**
- `removeProject()` now calls `cleanupProjectState(projectId, projectPath)` after removing the project from the list.

The cleanup functions are wrapped in try/catch inside the registry, so if one module isn't loaded or throws an error, it doesn't block the others.

## What I considered

**Cleaning up by project ID vs path.** FilesPanel uses project path as the key, while PreviewPanel uses project ID. The registry passes both to the cleanup function so each module can pick what it needs.

**Clearing the entire notifiedToolUseIds set.** Ideally we'd only remove IDs belonging to the deleted project, but the Set doesn't track which project each ID belongs to. Adding that tracking would be more complexity than it's worth for a notification dedup set. A full clear is cheap and safe.

**Module-level Maps vs stores.** The Maps are at module level by design — they need to survive component destruction. Moving them to a Svelte store wouldn't solve the cleanup problem, it would just move it somewhere else.

## Validation

- `bun run lint` — clean
- No new tests. This is infrastructure-level cleanup code. Testing it properly would require mocking the component lifecycle and the registry, which adds more code than the feature itself. The cleanup functions themselves are one-liners (`map.delete(key)`) that are hard to get wrong.
- Manual review of the call chain: `removeProject()` → `cleanupProjectState()` → each registered function.

### DIFF
--- a/frontend/components/workspace/panels/FilesPanel.svelte
+++ b/frontend/components/workspace/panels/FilesPanel.svelte
@@ -1,4 +1,6 @@
 <script module lang="ts">
+	import { registerProjectCleanup } from '$frontend/utils/project-state-cleanup';
+
 	// In-memory snapshot survives component destruction within the same SPA
 	// session (e.g. mobile/desktop layout switch). DB remains the source of
 	// truth for cross-device persistence.
@@ -23,13 +25,14 @@
 		projectFileStates.delete(projectPath);
 	}
 
-	/**
-	 * Clean up all persisted state.
-	 * Useful for testing or full reset scenarios.
-	 */
-	export function cleanupAllProjectStates(): void {
-		projectFileStates.clear();
+	function cleanupProjectFilePanelState(_projectId: string, projectPath?: string): void {
+		if (projectPath) {
+			cleanupProjectState(projectPath);
+		}
 	}
+
+	// Register once at module load to avoid duplicate closures on remount.
+	registerProjectCleanup(cleanupProjectFilePanelState);
 </script>
 
 <script lang="ts">
@@ -55,14 +58,6 @@
 		refreshGitStatus,
 		syncGitStatusForProject
 	} from '$frontend/stores/features/git-status.svelte';
-	import { registerProjectCleanup } from '$frontend/utils/project-state-cleanup';
-
-	// Register cleanup with the centralized registry
-	registerProjectCleanup((_projectId, projectPath) => {
-		if (projectPath) {
-			cleanupProjectState(projectPath);
-		}
-	});
 
 	// Props
 	interface Props {
--- a/frontend/components/workspace/panels/FilesPanel.svelte
+++ b/frontend/components/workspace/panels/FilesPanel.svelte
@@ -14,6 +14,22 @@
 	}
 
 	const PANEL_STATE_VERSION = 1;
+
+	/**
+	 * Clean up persisted state for a specific project.
+	 * Called when a project is removed to prevent memory leaks.
+	 */
+	export function cleanupProjectState(projectPath: string): void {
+		projectFileStates.delete(projectPath);
+	}
+
+	/**
+	 * Clean up all persisted state.
+	 * Useful for testing or full reset scenarios.
+	 */
+	export function cleanupAllProjectStates(): void {
+		projectFileStates.clear();
+	}
 </script>
 
 <script lang="ts">
@@ -39,6 +55,14 @@
 		refreshGitStatus,
 		syncGitStatusForProject
 	} from '$frontend/stores/features/git-status.svelte';
+	import { registerProjectCleanup } from '$frontend/utils/project-state-cleanup';
+
+	// Register cleanup with the centralized registry
+	registerProjectCleanup((_projectId, projectPath) => {
+		if (projectPath) {
+			cleanupProjectState(projectPath);
+		}
+	});
 
 	// Props
 	interface Props {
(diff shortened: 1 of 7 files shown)


ArgaFairuz:
Thanks @DeryFerd — the instinct to centralize cleanup behind a registry rather than scatter `delete` calls across the deletion path is the right shape, and the writeup that names the tradeoffs upfront (id-vs-path keying, dedup set scope, why the Maps stay at module level) made the audit easy to follow. A few things I'd like to talk through before merging.

The registration at `FilesPanel.svelte:61` and `PreviewPanel.svelte:40` sits inside the instance `<script lang="ts">`, not `<script module>`. Each mount adds a fresh closure to the `Set<CleanupFn>` — the Set dedupes by reference, but each closure is a new identity, so the registry grows without bound, and there's no `unregister` or `onDestroy` to clear it. The scenario these module-level Maps exist for (mobile/desktop layout switch, where components mount and unmount) is the same scenario that grows the registry. I'd update my position here if these panels are guaranteed to mount only once per session — otherwise the fix re-introduces a leak one layer up. Moving the `registerProjectCleanup` call into a `<script module>` block, or registering a single named module-level function reference, closes it.

A related premise check: `projectPreviewState` in `PreviewPanel.svelte:15` is declared in the instance `<script lang="ts">`, not `<script module>` — so it dies on unmount rather than persisting like `projectFileStates`. The PR description treats both Maps the same, but the lifetimes differ. Worth confirming the leak you observed was actually from this Map and not from somewhere upstream.

I also noticed an adjacent surface the PR didn't address: `frontend/services/terminal/project.service.ts:26` — `projectContexts` is a `Map<projectId, ProjectTerminalContext>` holding per-project session outputs and command histories, never cleaned up on `removeProject`. Same shape as the three sites here, and by data volume it's the larger of the leaks.

If you'd rather keep this PR scoped to the original three sites, happy to take the `projectContexts` cleanup as a follow-up myself and credit this PR for the pattern.

Two small things while you're in there: `project-state-cleanup.ts:33` uses `console.error`, but CLAUDE.md requires the `debug` module (`debug.error('cleanup', ...)`); and `cleanupAllProjectStates` / `cleanupAllProjectPreviewStates` are unused — safe to drop unless something is queued up that needs them.

Could you take a look by May 25, 2026? If you can't respond by then, I'll close this PR as auto-stale per CONTRIBUTING.md → After You Submit — you can reopen anytime once you're back.

DeryFerd:
Thanks for the deep review. I pushed a follow-up commit that addresses the lifecycle and registration issues you flagged.

What changed:
- Moved Files panel cleanup registration out of instance `<script>` and into a single module-level function registration, so remounts do not add new closure identities.
- Moved Preview panel per-project state into a dedicated module-level utility (`frontend/utils/preview-project-state.ts`) so state persists across panel unmount/remount and cleanup registration happens once.
- Removed the now-unneeded `cleanupAll*` helper exports in those panels (the per-project cleanup path is the active path now).
- Replaced `console.error` in centralized cleanup with the project logger (`debug.error('project', ...)`) to match repository logging conventions.
- Addressed the adjacent terminal context leak by adding per-project context teardown in `terminalProjectManager.cleanupProjectContext()` and wiring it into the same project cleanup registry.

Validation run:
- `bun run check`
- `bunx eslint frontend/utils/project-state-cleanup.ts frontend/utils/preview-project-state.ts frontend/services/terminal/project.service.ts frontend/services/notification/global-stream-monitor.ts`

ArgaFairuz:
Thanks @DeryFerd — this is exactly the shape I was hoping for. Moving `projectPreviewState` into a dedicated module-level utility at `frontend/utils/preview-project-state.ts` is cleaner than keeping it inline in the panel, the FilesPanel registration now sits in `<script module>` with a named function reference so remounts don't grow the registry, and the terminal context teardown at `frontend/services/terminal/project.service.ts:727` covers the adjacent surface I flagged — nicely scoped to listeners + persistence + session manager + the map itself. `bun run check` and `bun run lint` green locally. Merging.

One more adjacent gap I should have caught in my first review (not on you to chase): `frontend/components/files/FileTree.svelte:3` has the same shape — a module-level `projectSearchStates` Map keyed by project ID, populated on project switch (lines 104, 122, 162), never cleaned up on remove. I'll open a small follow-up myself (`fix/memory-cleanup-file-tree-search-state`) wiring it into the same registry and credit this PR for the pattern.

PR #256, opened 2026-05-17
fix(status): purge presence map when project is fully deleted

DeryFerd:
## What was going wrong

Collaboration presence in Clopen lives in a process-local `Map` inside `backend/project/status-manager.ts`, not in SQLite. When someone triggers a **full** project delete (`projects:delete` with `mode: 'full'`) and they're the last member, the handler already tears down sessions, chat streams, terminal streams, and eventually the project row. The presence map did not get the same treatment.

`updateUserPresence` only drops a project key when the last user explicitly leaves or when the five-minute idle sweep runs. After a full delete, `getProjectStatusData(projectId)` could still return `activeUsers` for a project id that no longer exists in the database. That mostly shows up as confusing collaboration/status UI—stale names on a ghost project—not as a security hole, but it's inconsistent with how aggressively we clean up everything else on delete.

## What this PR does

Adds a small, explicit `clearProjectPresence(projectId)` that deletes the map entry for that project. It runs in `backend/ws/projects/crud.ts` only in the narrow case we care about: full delete, zero remaining users, immediately before `projectQueries.deleteProject`.

No new WebSocket events, no change to join/leave flows, and no behavior change for `mode: 'remove'` (project row kept for restore).

## Files touched

| File | Change |
|------|--------|
| `backend/project/status-manager.ts` | New `clearProjectPresence()` |
| `backend/ws/projects/crud.ts` | Call it on orphaned full delete |
| `backend/project/status-manager.test.ts` | Join → clear → assert empty `activeUsers`; no-op on unknown id |

## How I checked it

- `bun test backend/project/status-manager.test.ts`
- `bun run check`
- `bun run lint`

I did not run the full app for this one; the bug is entirely in-memory state tied to a code path we can assert in unit tests. Happy to do a manual delete pass if a maintainer wants eyes on the collaboration panel.

## Merge note

Targets **`main`** directly. Independent of the cleanup-registry work in #255. If that registry lands later, we *could* also register `clearProjectPresence` there—but the `crud.ts` call alone is sufficient and avoids wiring the same cleanup twice unless someone deliberately wants a single registry entry.

### DIFF
--- a/backend/project/status-manager.test.ts
+++ b/backend/project/status-manager.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, test } from 'bun:test';
+import {
+	clearProjectPresence,
+	getProjectStatusData,
+	updateUserPresence
+} from './status-manager';
+
+async function getSingleProjectStatus(projectId: string) {
+	const status = await getProjectStatusData(projectId);
+	if (Array.isArray(status)) {
+		throw new Error('expected single-project status payload');
+	}
+	return status;
+}
+
+describe('clearProjectPresence', () => {
+	test('removes active users for the project from status data', async () => {
+		updateUserPresence('proj-presence', 'user-1', 'Alice', 'join');
+
+		const before = await getSingleProjectStatus('proj-presence');
+		expect(before.activeUsers).toHaveLength(1);
+
+		clearProjectPresence('proj-presence');
+
+		const after = await getSingleProjectStatus('proj-presence');
+		expect(after.activeUsers).toEqual([]);
+	});
+
+	test('is a no-op when the project has no presence entry', () => {
+		expect(() => clearProjectPresence('proj-never-seen')).not.toThrow();
+	});
+});
--- a/backend/project/status-manager.ts
+++ b/backend/project/status-manager.ts
@@ -176,6 +176,13 @@ export async function getProjectStatusData(projectId?: string) {
   }
 }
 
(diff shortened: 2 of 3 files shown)


ArgaFairuz:
Thanks @DeryFerd — clean, narrow fix and the call site placement is exactly right. The choice to exclude `mode: 'remove'` is the load-bearing detail here and you got it: presence has to survive partial removals because other users may still be active on the row.

One housekeeping note — this conflicted with main because #253 (`disposeProjectEngines`) landed in the same `if (remainingUsers === 0)` block at `backend/ws/projects/crud.ts`. Resolution was trivial — both cleanups now coexist before `deleteProject`, with engine disposal first and presence purge right after. Pushed back to your fork.

Walked the scenario end-to-end: two users on `proj-X`, one leaves, the other does a `mode: 'full'` delete — without the fix, `getProjectStatusData()` (`backend/project/status-manager.ts:141-158`) still synthesizes a `{ projectId, activeUsers }` entry for the deleted row until the 5-minute idle sweep at `:52` clears it. With the fix, map entry is dropped at the same instant as the DB row. Test coverage in `status-manager.test.ts` pins the function shape directly; sufficient for in-memory state of this size.

`bun run check` / `bun run lint` / `bun test backend/project/status-manager.test.ts` green after the merge. Will squash-merge once the push lands.

PR #258, opened 2026-05-17
fix(ws): add connection-level message rate limiting to prevent DoS

DeryFerd:
## Context

The WebSocket router currently has no guard against message spam from authenticated connections. Once a user logs in, they can fire unlimited messages per second — triggering file writes, git operations, chat streams, and terminal commands without any throttling. A compromised session or malicious insider could saturate the backend by flooding the connection.

## What This Adds

A per-connection rate limiter that sits between the auth gate and the route handler. It tracks message frequency using a sliding window and applies three tiers of response:

- **50 msg/s** — logs a warning, no action taken
- **100 msg/s** — starts dropping messages silently
- **200 msg/s** — flags the connection for potential disconnect (external handler decides whether to close)

The limiter uses a WeakMap keyed on the raw Bun ServerWebSocket, so state is automatically garbage-collected when connections close. No manual cleanup needed beyond a periodic tick that logs a heartbeat.

## Why These Thresholds

I picked 50/100/200 based on normal usage patterns. A typical user interacting with the frontend sends maybe 5-15 messages per second during active work (typing in chat, browsing files, switching projects). Even power users running batch file operations rarely exceed 30 msg/s. The warning tier at 50 gives headroom before any intervention, while the throttle at 100 catches sustained spam without affecting legitimate bursts.

The disconnect flag at 200 is intentionally non-automatic — the rate limiter only marks the connection. The actual disconnect decision stays with the server, so we don't accidentally kill someone on a flaky connection that's retrying messages.

## Implementation Details

The rate limiter plugs into `WSRouter` via a new `setRateLimiter()` method, mirroring the existing `setAuthMiddleware()` pattern. This means:

1. Auth gate runs first (blocks unauthenticated requests)
2. Rate limiter runs second (drops spam from authenticated connections)
3. Route handler runs last (only if both gates pass)

Messages that fail the rate check are silently ignored — no error response sent back. This is intentional: responding to spam gives the attacker a feedback channel. Silence is cheaper for them to deal with.

## Files Changed

- `backend/ws/message-rate-limiter.ts` — New module with `MessageRateLimiter` class
- `shared/utils/ws-server.ts` — Added `setRateLimiter()` and rate limit gate in `handleMessage()`
- `shared/utils/logger.ts` — Added `'rate-limit'` to the `LogLabel` type
- `backend/index.ts` — Registered the limiter on wsRouter and added disposal on shutdown

## Verification

- `bun tsc --noEmit` passes clean
- No breaking changes — the rate limiter is additive and does nothing when under thresholds
- Disposal on shutdown prevents timer leaks

### DIFF
--- a/backend/index.ts
+++ b/backend/index.ts
@@ -39,12 +39,18 @@ import { checkRouteAccess } from './auth/permissions';
 import { authRateLimiter } from './auth';
 import { sessionCleanupScheduler } from './auth/session-cleanup';
 import { ws as wsServer } from './utils/ws';
+import { messageRateLimiter } from './ws/message-rate-limiter';
 
 // Register auth gate on WebSocket router — blocks unauthenticated/unauthorized access
 wsRouter.setAuthMiddleware(async (conn, action) => {
-	const isAuth = wsServer.isAuthenticated(conn);
-	const role = wsServer.getRole(conn);
-	return checkRouteAccess(action, isAuth, role);
+  const isAuth = wsServer.isAuthenticated(conn);
+  const role = wsServer.getRole(conn);
+  return checkRouteAccess(action, isAuth, role);
+});
+
+// Register message rate limiter on WebSocket router — prevents DoS via message spam
+wsRouter.setRateLimiter((conn, action) => {
+  return messageRateLimiter.checkRateLimit(conn, action);
 });
 
 /**
@@ -192,6 +198,8 @@ async function gracefulShutdown() {
 		app.stop();
 		// Dispose rate limiter timer
 		authRateLimiter.dispose();
+		// Dispose message rate limiter timer
+		messageRateLimiter.dispose();
 		// Dispose expired session cleanup timer
 		sessionCleanupScheduler.dispose();
 		// Close MCP remote server (before engines, as they may still reference it)
--- a/backend/index.ts
+++ b/backend/index.ts
@@ -43,14 +43,14 @@ import { messageRateLimiter } from './ws/message-rate-limiter';
 
 // Register auth gate on WebSocket router — blocks unauthenticated/unauthorized access
 wsRouter.setAuthMiddleware(async (conn, action) => {
-  const isAuth = wsServer.isAuthenticated(conn);
-  const role = wsServer.getRole(conn);
-  return checkRouteAccess(action, isAuth, role);
+	const isAuth = wsServer.isAuthenticated(conn);
+	const role = wsServer.getRole(conn);
+	return checkRouteAccess(action, isAuth, role);
 });
 
 // Register message rate limiter on WebSocket router — prevents DoS via message spam
 wsRouter.setRateLimiter((conn, action) => {
-  return messageRateLimiter.checkRateLimit(conn, action);
+	return messageRateLimiter.checkRateLimit(conn, action);
 });
 
 /**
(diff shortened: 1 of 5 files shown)


ArgaFairuz:
Thanks @DeryFerd — the threat model in the description is clear, and mirroring the `setAuthMiddleware` shape on `WSRouter` is exactly the right pattern fit. Two design questions before I can take this through, plus a few smaller items worth sweeping in the same revision.

At `backend/ws/message-rate-limiter.ts:50-55`, the disconnect tier (`>=200 msg/s`) sets `isFlagged` but returns `true` — so the message still reaches the route handler. The description says "external handler decides whether to close", but I can't find a consumer of `isFlaggedForDisconnect()` in the diff. Is that consumer landing in a follow-up, or is the intent that 200+ msg/s should drop the message *and* schedule a close? If the latter, the disconnect branch should return `false` and the close path should be wired here. I'd update my position if there's a downstream PR meant to consume the flag.

[CONTRIBUTING.md → Tests](./CONTRIBUTING.md#tests) expects a `*.test.ts` for security-class changes — at minimum a case that fires 100 messages and asserts the gate drops them, plus a happy path under 50 msg/s. The sliding-window math at `checkRateLimit` is exactly the branching that's silent when it regresses, so a `message-rate-limiter.test.ts` alongside the source would pin it.

Smaller items worth folding in:

- `message-rate-limiter.ts:118-120` — `globalCleanup()` only logs a heartbeat, and the description confirms no cleanup is needed (WeakMap GCs on close). The 30s interval + `dispose()` can both go.
- `message-rate-limiter.ts:74` — the warning tier calls `debug.warn` on every message in the band, so a sustained 60 msg/s connection floods 60 log lines/s. The throttle/disconnect tiers use a transition latch (`if (!state.isThrottled)`) — the same shape here would log once per entry.
- `backend/index.ts:46-53` and `shared/utils/ws-server.ts:283-296,628-634` use 2-space indent inside a tab-indented file. [CONTRIBUTING.md → Formatting](./CONTRIBUTING.md#formatting) is tabs; lint doesn't catch it because there's no Prettier.

Could you take a look by May 25, 2026? If you can't respond by then, I'll close this PR as auto-stale per [After You Submit](./CONTRIBUTING.md#after-you-submit) — you can reopen anytime once you're back.

DeryFerd:
Thanks for the detailed review. I pushed a follow-up commit to address the requested changes:

- Disconnect tier (`>=200 msg/s`) now drops the triggering message and immediately closes the connection (`1008`) from the limiter path.
- Added regression tests for the limiter thresholds (`<50` pass, `>=100` drop, `>=200` close + drop): `backend/ws/message-rate-limiter.test.ts`.
- Removed the no-op global cleanup interval and the message-rate-limiter `dispose()` wiring from shutdown.
- Reduced warning-tier log spam by logging only on threshold-band transition (same transition-latch shape as throttle/disconnect).
- Normalized tab indentation in the touched sections of `backend/index.ts` and `shared/utils/ws-server.ts`.

Validation run:
- `bun test backend/ws/message-rate-limiter.test.ts`
- `bunx eslint backend/ws/message-rate-limiter.ts backend/ws/message-rate-limiter.test.ts backend/index.ts shared/utils/ws-server.ts`

ArgaFairuz:
Thanks @DeryFerd — fast turnaround, and you caught the things that mattered: the `1008` close wiring on the disconnect tier, the transition-latch on warning, and a clean mock-based test for each threshold. The isolation test using `throttle=300, disconnect=200` to pin the close path without throttle drops interfering first was a good call.

Built on top with one mechanical pass: normalized tab indentation in `backend/ws/message-rate-limiter.ts` (the `interface` and `DEFAULT_CONFIG` block, the opening of `checkRateLimit`, and the trailing helper methods were still 2-space) and `shared/utils/ws-server.ts` (the comment block at `:298` had regressed to 2-space, and the new `RATE LIMIT GATE` block at `:628-634` sat one tab deeper than the `AUTH GATE` start at `:602`). `bun run check` / `bun run lint` / `bun test` green, synced with `main`.

PR #259, opened 2026-05-17
fix(engine): validate MCP tool output before forwarding to AI models

DeryFerd:
## The Problem

MCP tools run in-process and return arbitrary data from the filesystem, browser automation, or external APIs. There's currently no size check on tool outputs before they get forwarded to AI models. A tool that reads a large file, captures a high-res screenshot, or returns verbose debug data can push a single response past tens of megabytes.

This matters because the AI engine adapters (Codex, Claude, Copilot, Qwen, OpenCode) serialize these responses into their message streams. A single oversized tool output can cause memory pressure, slow down stream processing, or hit upstream API limits that the engine doesn't handle gracefully.

## The Fix

Added an output validator that runs inside `createRemoteMcpServer()` — the function that registers all MCP tools for the remote HTTP transport. Every tool handler's result passes through `validateMcpOutput()` before being returned to the engine.

The validator checks three limits:

| Content type | Limit | Action when exceeded |
|---|---|---|
| Text | 10 MB | Truncated with ellipsis and marker |
| Image (base64) | 5 MB | Replaced with text explaining what was omitted |
| Total per tool call | 15 MB | Remaining content dropped with marker |

Size estimation uses conservative heuristics: 2 bytes per character for text (UTF-8 worst case), and 75% of base64 string length for images (since base64 expands data by ~33%).

## Why These Limits

10 MB for text is generous — most tool outputs are under 100 KB. Even a full file listing of a large project rarely exceeds 1 MB. The 10 MB ceiling catches pathological cases (reading binary files as text, dumping entire logs) without blocking legitimate large outputs.

5 MB for images covers typical screenshots at reasonable resolutions. A 1920x1080 screenshot compressed as PNG is usually 200-500 KB. Even at 4K, you'd need a very detailed image to hit 5 MB.

15 MB total prevents the "death by a thousand cuts" scenario where many medium-sized outputs combine into a massive response.

## Edge Cases Handled

- Empty content arrays pass through unchanged
- Non-text/non-image content types (if any are added later) pass through without modification
- Truncation markers are appended so the AI model knows data was cut — it won't silently assume the output was complete
- The validator runs after the abort-signal check, so cancelled tools don't waste cycles on validation

## Files Changed

- `backend/mcp/output-validator.ts` — New module with `validateMcpOutput()` and `getMcpOutputLimits()`
- `backend/mcp/servers/helper.ts` — Wrapped tool handler output in `createRemoteMcpServer()` with validation

## Verification

- `bun tsc --noEmit` passes clean
- Applies to all MCP tools registered via the remote HTTP transport (browser automation, weather, and any future tools)
- Does not affect Claude Code's in-process MCP path (`createSdkMcpServer`) — that path uses the same tool definitions but a different registration flow. If validation is needed there too, it should be added at the SDK level or in a shared wrapper.

### DIFF
--- a/backend/mcp/config.ts
+++ b/backend/mcp/config.ts
@@ -15,6 +15,7 @@ import { serverRegistry, serverFactories, serverMetadata } from './servers';
 import { projectContextService } from './project-context';
 import { debug } from '$shared/utils/logger';
 import { SERVER_ENV } from '../utils/env';
+import { validateMcpOutput } from './output-validator';
 
 /**
  * User-defined MCP Servers Configuration
@@ -117,7 +118,9 @@ export function getEnabledMcpServers(context?: McpExecutionContext): Record<stri
 							if (signal?.aborted) {
 								return abortedToolResult(toolName);
 							}
-							return def.handler(args);
+							const result = await def.handler(args);
+							result.content = validateMcpOutput(result.content, toolName);
+							return result;
 						});
 					};
 					return tool(toolName, def.description, def.schema, boundHandler as any);
--- a/backend/mcp/output-validator.test.ts
+++ b/backend/mcp/output-validator.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, test } from 'bun:test';
+import { validateMcpOutput } from './output-validator';
+
+const MB = 1024 * 1024;
+
+describe('validateMcpOutput', () => {
+	test('truncates oversized text output to the configured text budget', () => {
+		const oversizedText = 'a'.repeat((10 * MB) + 8_192);
+
+		const result = validateMcpOutput([{ type: 'text', text: oversizedText }], 'tool-a');
+
+		expect(result).toHaveLength(1);
+		expect(result[0].type).toBe('text');
+		expect(result[0].text).toContain('[Content truncated due to size limit]');
+		expect(new TextEncoder().encode(result[0].text ?? '').length).toBeLessThanOrEqual(10 * MB);
+	});
+
+	test('replaces oversized image output with explanatory text', () => {
+		const oversizedImage = 'a'.repeat(Math.ceil((6 * MB) / 0.75));
+
+		const result = validateMcpOutput([
+			{ type: 'image', data: oversizedImage, mimeType: 'image/png' }
+		], 'take_screenshot');
+
+		expect(result).toEqual([
+			{
+				type: 'text',
+				text: expect.stringContaining('[Image omitted: size')
+			}
+		]);
+	});
+
+	test('stops once sanitized total output exceeds the overall cap', () => {
+		const elevenMbText = 'b'.repeat(11 * MB);
+		const sixMbText = 'c'.repeat(6 * MB);
+		const trailingText = 'should never appear';
+
+		const result = validateMcpOutput([
+			{ type: 'text', text: elevenMbText },
+			{ type: 'text', text: sixMbText },
+			{ type: 'text', text: trailingText }
+		], 'tool-b');
+
+		expect(result).toHaveLength(2);
+		expect(result[0].text).toContain('[Content truncated due to size limit]');
+		expect(result[1]).toEqual({
+			type: 'text',
+			text: '\n\n[Additional content omitted due to total size limit]'
+		});
+		expect(result.some(item => item.text === trailingText)).toBe(false);
+	});
+});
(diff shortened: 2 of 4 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks for this — the failure modes you walked through in the description (memory pressure, upstream API limits, the heuristic justification for each limit) are exactly the shape of audit notes that make this easy to review. One coverage question and a couple of smaller items before we can merge.

## Coverage asymmetry

The validator only runs on the remote-HTTP path at `backend/mcp/servers/helper.ts:208`, which serves Open Code / Codex / Copilot. The Claude Code in-process path at `backend/mcp/config.ts:120` invokes the same `def.handler(args)` — same `RawToolDef`, same handler reference — but bypasses validation entirely. Concretely: `take_screenshot` returns base64 from `backend/mcp/servers/browser-automation/inspection.ts:463`, and a full-page 4K screenshot hitting Claude wouldn't be capped.

You flagged this in the description ("should be added at the SDK level or in a shared wrapper"), but the wrapping site already exists — `boundHandler` at `backend/mcp/config.ts:114-122` is the symmetric spot, and `result.content = validateMcpOutput(...)` slots in with the same shape as your `helper.ts` change. Is there a reason to keep the asymmetry, or would you extend coverage in this PR? If the SDK already enforces a server-side limit on Claude's side I'd want to see that documented; otherwise the title's "before forwarding to AI models" reads as covering both engines.

## Please add a regression test

The validator has three branches (text truncation, image-omitted-with-text-replacement, total-cap break) plus the truncate helper's character-budget math — all pure functions of `(content, toolName)`, easy to pin. `backend/mcp/` doesn't have an existing test directory but `*.test.ts` colocated next to `output-validator.ts` matches the repo's convention elsewhere (per CONTRIBUTING.md → Tests).

## Minor (optional)

- `getMcpOutputLimits()` at `backend/mcp/output-validator.ts:85` is exported but unused — either drop it or add the caller you had in mind.
- `totalSize` accumulates pre-truncation sizes (`output-validator.ts:54, 64`), so once text is truncated to 10 MB the total-cap still counts the original size against the 15 MB budget. Not a bug, but the "15 MB total" docstring slightly overstates the effective cap.
- `text.length * 2` for UTF-8 estimation is safely conservative; `new TextEncoder().encode(text).length` is one line for the actual count if you want to drop the 2× safety factor.

Could you take a look by May 25, 2026? If you can't respond by then, I'll close this PR as auto-stale per CONTRIBUTING.md → After You Submit — reopen anytime once you're back.

DeryFerd:
Thanks for the review. I pushed a follow-up commit that closes the coverage gap and adds regression tests.

What changed:
- Added the same `validateMcpOutput(...)` wrapper to the Claude/in-process path in `backend/mcp/config.ts`, so both the in-process SDK path and the remote-HTTP helper path now validate tool output before returning it to the engine.
- Added colocated regression tests in `backend/mcp/output-validator.test.ts` covering oversized text truncation, oversized image replacement, and the total-budget early-stop branch.
- Dropped the unused `getMcpOutputLimits()` export.
- Tightened the validator so total-budget accounting uses the sanitized payload size rather than the original pre-truncation estimate, and switched text-size estimation to `TextEncoder` for an exact UTF-8 byte count.

Validation run:
- `bun test backend/mcp/output-validator.test.ts`
- `bunx eslint backend/mcp/output-validator.ts backend/mcp/output-validator.test.ts backend/mcp/config.ts backend/mcp/servers/helper.ts`
- `bun run check`

ArgaFairuz:
Thanks @DeryFerd — clean follow-up across all four items. Lifting `validateMcpOutput` into the `boundHandler` at `backend/mcp/config.ts:118` is the symmetric spot, and the colocated test in `backend/mcp/output-validator.test.ts` pins all three branches (text truncation, image-omitted-with-text, total-cap break) so the next refactor can't regress quietly. Switching `estimateTextSize` to `TextEncoder.encode().length` removes the 2× safety margin without losing the guarantee, and totalling from the sanitized item at `backend/mcp/output-validator.ts:75,83` makes the effective 15 MB cap match the docstring.

One small thing before squashing: the follow-up commit's `fix(mcp): validate tool output across all engine paths` reads truer to the broader scope than the current PR title `fix(engine): ...`. Worth retitling the PR if you'd rather the squash subject reflect that the in-process path is now covered too.

`bun run check` / `lint` / `test` green locally. Merging.

PR #260, opened 2026-05-17
feat(auth): add session activity logging for audit trail

DeryFerd:
## Problem

Right now, authentication events happen in memory with no persistent record. When a user logs in, gets logged out, has project access revoked, or an admin kicks everyone out — there's no trail to look back on. If someone reports "I was logged out unexpectedly" or "I lost access to a project," there's no way to investigate what happened.

This is especially relevant now that PR #250 added session invalidation on project access revocation. Without logging, admins can't see who was affected by an invalidation event or verify that it actually fired.

## What Gets Tracked

Every entry in the audit log records: who, what, when, and optional context (IP address, details string).

| Event type | Triggered when | Details captured |
|---|---|---|
| `auth:login` | Successful login via PAT or session token | User name, token type, IP address |
| `auth:logout` | User clicks logout | IP address |
| `auth:logout-all` | Admin kicks all sessions | Admin user ID, session count cleared |
| `project:assigned` | Admin grants project access to a user | User ID, project ID |
| `project:unassigned` | Admin revokes project access | User ID, project ID (this is where PR #250's invalidation fires) |

## The Database Side

Migration 036 creates `auth_audit_log` with three indexes:

- `idx_auth_audit_user_id` — query by user (useful for investigating a specific account)
- `idx_auth_audit_event_type` — query by event type (useful for finding all logins, or all project unassignments)
- `idx_auth_audit_created_at` — query by date range (useful for time-bounded investigations)

The `id` column uses `randomblob(16)` for UUIDs, consistent with the rest of the schema.

## Query Module

`auditLogQueries` provides five accessors plus a cleanup function:

- `logEvent()` — insert a new entry
- `getUserLogs(userId)` — get recent events for a specific user
- `getEventsByType(eventType)` — get recent events of a specific type
- `getRecentLogs()` — get the latest events across all users (admin dashboard fodder)
- `getLogsByDateRange(start, end)` — time-bounded query
- `deleteOldLogs(beforeDate)` — cleanup old entries to prevent table bloat

None of these are exposed via API yet. They're available for backend use — an admin dashboard or CLI tool could wire them up later.

## Files Changed

- `backend/database/migrations/036_create_auth_audit_log.ts` — New migration
- `backend/database/queries/audit-log-queries.ts` — Query module
- `backend/database/queries/index.ts` — Export `auditLogQueries`
- `backend/database/migrations/index.ts` — Register migration 036
- `backend/auth/auth-service.ts` — Audit logging on project assign/unassign
- `backend/ws/auth/login.ts` — Audit logging on login, logout, logout-all

## Notes

- Migration uses `CREATE TABLE IF NOT EXISTS` — safe to run on existing databases
- Logging is synchronous and in-process. If this becomes a bottleneck at scale, the `logEvent()` function could be made async or batched
- IP addresses come from `ws.getRemoteAddress()`, which returns the raw socket address. Behind a reverse proxy, this would need X-Forwarded-For handling

### DIFF
--- a/backend/auth/auth-service.ts
+++ b/backend/auth/auth-service.ts
@@ -524,7 +524,6 @@ export function unassignProjectFromUser(userId: string, projectId: string): bool
 	}
 	projectQueries.removeUserProject(userId, projectId);
 	debug.log('auth', `Project ${projectId} unassigned from user ${userId}`);
-
 	return true;
 }
 
--- a/backend/database/migrations/036_create_auth_audit_log.ts
+++ b/backend/database/migrations/036_create_auth_audit_log.ts
@@ -1,0 +1,28 @@
+/**
+ * Migration 036: Create auth_audit_log table
+ */
+
+import type { DatabaseConnection } from '$shared/types/database/connection';
+
+export const description = 'Create auth audit log table';
+
+export function up(db: DatabaseConnection): void {
+  db.exec(`
+    CREATE TABLE IF NOT EXISTS auth_audit_log (
+      id TEXT PRIMARY KEY DEFAULT (lower(hex(randomblob(16)))),
+      user_id TEXT NOT NULL,
+      event_type TEXT NOT NULL,
+      event_details TEXT,
+      ip_address TEXT,
+      user_agent TEXT,
+      created_at TEXT NOT NULL DEFAULT (datetime('now'))
+    );
+    CREATE INDEX IF NOT EXISTS idx_auth_audit_user_id ON auth_audit_log(user_id);
+    CREATE INDEX IF NOT EXISTS idx_auth_audit_event_type ON auth_audit_log(event_type);
+    CREATE INDEX IF NOT EXISTS idx_auth_audit_created_at ON auth_audit_log(created_at);
+  `);
+}
+
+export function down(db: DatabaseConnection): void {
+  db.exec(`DROP TABLE IF EXISTS auth_audit_log;`);
+}
--- a/backend/database/migrations/036_create_auth_audit_log.ts
+++ b/backend/database/migrations/036_create_auth_audit_log.ts
@@ -7,22 +7,24 @@ import type { DatabaseConnection } from '$shared/types/database/connection';
 export const description = 'Create auth audit log table';
 
 export function up(db: DatabaseConnection): void {
-  db.exec(`
-    CREATE TABLE IF NOT EXISTS auth_audit_log (
-      id TEXT PRIMARY KEY DEFAULT (lower(hex(randomblob(16)))),
-      user_id TEXT NOT NULL,
-      event_type TEXT NOT NULL,
-      event_details TEXT,
-      ip_address TEXT,
-      user_agent TEXT,
-      created_at TEXT NOT NULL DEFAULT (datetime('now'))
-    );
-    CREATE INDEX IF NOT EXISTS idx_auth_audit_user_id ON auth_audit_log(user_id);
-    CREATE INDEX IF NOT EXISTS idx_auth_audit_event_type ON auth_audit_log(event_type);
-    CREATE INDEX IF NOT EXISTS idx_auth_audit_created_at ON auth_audit_log(created_at);
-  `);
+	db.exec(`
+		CREATE TABLE IF NOT EXISTS auth_audit_log (
+			id TEXT PRIMARY KEY DEFAULT (lower(hex(randomblob(16)))),
+			user_id TEXT NOT NULL,
+			actor_user_id TEXT,
+			event_type TEXT NOT NULL,
+			event_details TEXT,
+			ip_address TEXT,
+			user_agent TEXT,
+			created_at TEXT NOT NULL DEFAULT (datetime('now'))
+		);
+		CREATE INDEX IF NOT EXISTS idx_auth_audit_user_id ON auth_audit_log(user_id);
+		CREATE INDEX IF NOT EXISTS idx_auth_audit_actor_user_id ON auth_audit_log(actor_user_id);
+		CREATE INDEX IF NOT EXISTS idx_auth_audit_event_type ON auth_audit_log(event_type);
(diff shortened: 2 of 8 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks for the PR — the framing in the description is sharp (tying this to #250's session invalidation makes the motivation legible), and the index choices on `auth_audit_log` cover the three query shapes that actually matter for investigations. A few things I'd like to discuss before this can be merged.

## Blockers

### 1. Actor identity is missing on `project:assigned` / `project:unassigned`

The PR description names the investigative use case as *"If someone reports 'I lost access to a project,' there's no way to investigate what happened."* With the current shape, `auditLogQueries.logEvent({ userId, ... })` at `backend/auth/auth-service.ts:506` and `:528` records the **target user** — the person who lost access — but not the **admin who performed the revoke**. The audit log can answer *when* and *who was affected*, but not *who did it*, which is the half that actually closes the investigation.

Two shapes that would close this:

- Add an `actor_user_id` column to `auth_audit_log` and pass the admin's id through `assignProjectToUser` / `unassignProjectFromUser`. The admin id is available at the handler boundary (`backend/ws/auth/users.ts:62` and `:83`) via `ws.getUserId(conn)`.
- Or record two rows per event (one for the actor, one for the target). Slightly noisier but keeps the schema flatter.

I'd lean toward the first — it preserves the one-row-per-event shape and the indexes you already have. What would flip my position: if you have a reason to prefer the two-row shape (e.g. it makes a planned admin-dashboard query simpler), I'd want to hear it before deciding.

### 2. Indentation mixes tabs and 2-space within the same files

The codebase uses tabs (see `backend/database/queries/auth-queries.ts`, migrations 001–035). The new code introduces 2-space indent in:

- `backend/auth/auth-service.ts:504-507` and `:526-529` — four new lines per function in 2-space while the surrounding lines are tab-indented.
- `backend/ws/auth/login.ts:142-148`, `:232-244`, `:259-269` — modified blocks converted from tab to 2-space.
- The new files (`backend/database/migrations/036_create_auth_audit_log.ts`, `backend/database/queries/audit-log-queries.ts`) are 2-space throughout.

`bun run lint` passes because there's no indent rule configured, but the inconsistency is visible in the diff. Could you reformat to tabs to match the rest of the codebase?

## Please add before merge

### A regression test for `logEvent` and at least one accessor

Per [CONTRIBUTING.md → Tests](./CONTRIBUTING.md#tests), changes of this surface warrant a test. The pattern lives at `backend/auth/permissions.test.ts`. Minimum: one test asserting that `logEvent` inserts a row with the expected columns, plus one for `getUserLogs` or `getEventsByType`. Without it, a future schema drift or column rename can silently break the audit trail — which is exactly the kind of failure mode the audit log is supposed to protect against.

- `backend/database/migrations/index.ts:253` — the entry hardcodes `description: 'Create auth audit log table'` while every other migration uses `migration0XX.description`. One source of truth is cleaner.
- `user_agent` column is defined in the schema but never populated by any callsite. Either drop the column or capture it from the WS connection.
- `event_details` strings like `` `Project ${projectId} assigned to user ${userId}` `` embed data that already lives in its own column (and would live in `actor_user_id` after blocker 1). Future "all events touching project X" queries would need `LIKE` matching. Consider structured columns or a JSON payload.
- `logEvent` runs without try/catch. If the audit insert fails (disk full, schema drift), the user-facing login/logout/assign throws. Best-effort observability shouldn't be able to break the primary flow — wrap the call.
- `auth:logout-all` at `backend/ws/auth/login.ts:267` doesn't capture `ip_address`, but `auth:login` (line 146) and `auth:logout` (line 241) do. Small consistency nit.

Could you take a look by May 25, 2026? If you can't respond by then, I'll close this PR as auto-stale per [CONTRIBUTING.md → After You Submit](./CONTRIBUTING.md#after-you-submit) — you can reopen anytime once you're back.

DeryFerd:
Pushed a follow-up that addresses the blockers and rebases the branch onto the current main.

What changed:
- Added `actor_user_id` to `auth_audit_log` and threaded actor capture through the WS handler boundary for `project:assigned` / `project:unassigned`, so the audit trail now records both the affected user and the admin who performed the action.
- Moved the project assignment audit writes out of `auth-service.ts` and into `backend/ws/auth/users.ts`, which keeps actor/request context at the route layer.
- Normalized the touched files to tab indentation to match the rest of the backend.
- Added `backend/database/queries/audit-log-queries.test.ts` covering `logEvent`, `getUserLogs`, `getEventsByType`, and the best-effort failure path.
- Wrapped `logEvent` in a try/catch so audit insert failures do not break login/logout/project-assignment flows.
- Switched the migration index entry to `migration036.description` and added `ip_address` capture for `auth:logout-all`.

While rebasing onto the latest `main`, I kept the newer `invalidateUserSessions(...)` behavior in `auth:unassign-project` and combined it with the audit logging change.

Validation run:
- `bun test backend/database/queries/audit-log-queries.test.ts`
- `bunx eslint backend/auth/auth-service.ts backend/ws/auth/login.ts backend/ws/auth/users.ts backend/database/queries/audit-log-queries.ts backend/database/queries/audit-log-queries.test.ts backend/database/migrations/036_create_auth_audit_log.ts backend/database/migrations/index.ts`
- `bun run check`

ArgaFairuz:
Thanks @DeryFerd — actor capture at the route layer is exactly the shape I had in mind, and the best-effort try/catch around `logEvent` with the dedicated regression test closes the failure-mode gap cleanly. Built on top with one small fix.

While reconciling with #250's `invalidateUserSessions` change, the original call at `backend/ws/auth/users.ts:104` was kept in addition to the new one you placed at `:96`, so the function was firing twice per revoke. Not a crash (the second call finds 0 connections), but it leaks a misleading `"No active sessions to invalidate"` debug line on every legitimate revoke. Dropped the duplicate at `:104` so the order is now state change → invalidate → audit log → emit. `bun test backend/database/queries/audit-log-queries.test.ts` and `bun run check` / `bun run lint` green, synced with `main`.

PR #263, opened 2026-05-19
fix(files): guard symlink resolution in path access checks

DeryFerd:
### What & Why

The file access guards in `backend/ws/files/path-access.ts` rely on `resolve()` from Node's `path` module to normalize paths before checking if a file lives inside a project root. Problem is, `resolve()` doesn't follow symlinks. If someone on your team creates a symlink inside a project directory that points somewhere else—say `/home/project/link` → `/etc`—the current check sees the resolved path as `/etc` and blocks it. But what if the symlink is in a project that *another user has access to*? The access check passes because the symlink *file itself* is inside the project, even though `realpath` would resolve it to a completely different location.

Concretely: a member user with read/write access to a project could create a symlink pointing outside their sandbox, then read or write files through it. The access check at `isPathInside` only sees the normalized parent of the symlink, not where it actually leads.

This fix adds a `resolveRealPath()` wrapper that calls `realpath()` (which follows symlinks) before doing the containment check. If the path doesn't exist on disk, it falls back to `resolve()` so new-file operations still work.

### Changes

- `backend/ws/files/path-access.ts` — added `resolveRealPath()`, made `isPathInside`, `requireFilePathAccess`, `requireSharedFilePathAccess`, and `filterAccessibleExpandedPaths` async, swapped `resolve()` with `resolveRealPath()` in all containment checks.
- `backend/ws/files/read.ts` — added `await` to calls that are now async.
- `backend/ws/files/write.ts` — added `await` to calls that are now async.

### Why this is safe

- Paths that don't exist yet (new files) fall through to `resolve()` — no breaking change for create/write operations.
- Admin users already bypass the access check, so there's zero perf impact for them.
- The fallback is per-call, not cached — no stale symlink issues.

### Test notes

Any test that mocks `requireFilePathAccess` will need to handle the async return now. I didn't find existing tests that directly unit-test these guards; the file access tests seem to go through the WS handlers instead. Happy to add them if you'd like.

### DIFF
--- a/backend/ws/files/path-access.test.ts
+++ b/backend/ws/files/path-access.test.ts
@@ -37,7 +37,7 @@ mock.module('../../database/queries/project-queries', () => ({
 	}
 }));
 
-const mockGetRole = mock(() => 'member' as const);
+const mockGetRole = mock((): 'admin' | 'member' => 'member');
 const mockGetUserId = mock(() => 'user-1');
 
 mock.module('$backend/utils/ws', () => ({
--- a/backend/ws/files/path-access.test.ts
+++ b/backend/ws/files/path-access.test.ts
@@ -1,0 +1,163 @@
+import { describe, expect, test, mock, beforeEach, afterEach } from 'bun:test';
+import { mkdir, symlink, writeFile, rm, stat } from 'node:fs/promises';
+import { join } from 'node:path';
+import { tmpdir, platform } from 'node:os';
+import { randomUUID } from 'node:crypto';
+
+const isWindows = platform() === 'win32';
+
+// Symlinks on Windows require admin privileges. Skip symlink-specific tests
+// when running without elevated permissions.
+async function canCreateSymlinks(): Promise<boolean> {
+	const testPath = join(tmpdir(), `symlink-test-${randomUUID()}`);
+	const targetPath = join(tmpdir(), `symlink-target-${randomUUID()}`);
+	try {
+		await mkdir(targetPath, { recursive: true });
+		await symlink(targetPath, testPath, 'dir');
+		await rm(testPath);
+		await rm(targetPath);
+		return true;
+	} catch {
+		return false;
+	}
+}
+
+// Mock dependencies before importing the module under test
+const mockGetAllForUser = mock(() => [] as Array<{ id: string; path: string }>);
+const mockGetAll = mock(() => [] as Array<{ id: string; path: string }>);
+const mockUserHasProject = mock(() => false);
+const mockGetByPath = mock(() => null as { id: string; path: string } | null);
+
+mock.module('../../database/queries/project-queries', () => ({
+	projectQueries: {
+		getAllForUser: mockGetAllForUser,
+		getAll: mockGetAll,
+		userHasProject: mockUserHasProject,
+		getByPath: mockGetByPath,
+	}
+}));
+
+const mockGetRole = mock(() => 'member' as const);
+const mockGetUserId = mock(() => 'user-1');
+
+mock.module('$backend/utils/ws', () => ({
+	ws: {
+		getRole: mockGetRole,
+		getUserId: mockGetUserId,
+	}
+}));
+
+mock.module('../access', () => ({
+	requireProjectAccess: mock(() => ({ id: 'proj-1', path: '' })),
+}));
+
+// Import after mocking
+const { requireFilePathAccess, requireSharedFilePathAccess } = await import('./path-access');
+
+let testDir: string;
+
+describe('path-access symlink resolution', () => {
+	beforeEach(async () => {
+		testDir = join(tmpdir(), `clopen-path-access-test-${randomUUID()}`);
+		await mkdir(testDir, { recursive: true });
+		mockGetRole.mockImplementation(() => 'member');
+		mockGetUserId.mockImplementation(() => 'user-1');
+	});
+
+	afterEach(async () => {
+		await rm(testDir, { recursive: true, force: true });
+	});
+
+	test('blocks read through symlink to external directory', async () => {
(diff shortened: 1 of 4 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks for spotting this — the writeup is unusually clear (named the threat model up front, walked through the access-via-symlink scenario in user-visible terms, and proactively flagged the async-test consequence). Before this can land, a couple of issues need to be addressed.

## Blockers

### 1. Logic inversion in `requireSharedFilePathAccess` ancestor guard

At `backend/ws/files/path-access.ts:79` (new diff), the patch drops the `!` from the ancestor-guard check:

```diff
-		if (!isPathInside(normalizedPath, project.path)) continue;
+		if (await isPathInside(normalizedPath, project.path)) continue;
```

This inverts the loop. Original: skip projects that are *not* inside `normalizedPath`, then reject if the user can't access the remaining (descendant) projects. Patched: skip projects that *are* inside `normalizedPath` — exactly the ones the guard exists to check. The ancestor protection is silently disabled.

**Repro:** user A has a project at `/users/a/project`; user B has no access to it. B issues `rename /users/a -> /users/x` via FolderBrowser. Before this PR: `Access denied`. After this PR: the loop skips A's project at the inverted check and the rename proceeds, clobbering A's project root. I verified `main` carries the `!` (`path-access.ts:62`) and the regression is new to this branch.

### 2. Symlink bypass still reachable for non-existing leaves

`resolveRealPath()` falls back to `resolve()` whenever `realpath` throws, which it does for any path whose leaf doesn't exist on disk yet. That covers legitimate create/write flows — but it also covers an attacker-controlled `dirname` with a non-existing `basename`, so a symlinked parent directory is never followed.

**Repro:** member user with read/write on `/proj` creates `ln -s /etc /proj/escape`. The PR correctly blocks `read /proj/escape` (realpath returns `/etc`, fails `isPathInside`). But `write /proj/escape/passwd`:

1. `realpath('/proj/escape/passwd')` throws `ENOENT` (`passwd` doesn't exist).
2. Fallback `resolve('/proj/escape/passwd')` returns the literal path — `escape` is never resolved.
3. `isPathInside('/proj', '/proj/escape/passwd')` returns true.
4. The downstream `Bun.write` follows the symlink and writes to `/etc/passwd`.

The fix shape is to realpath the deepest existing ancestor and rejoin the unresolved tail, so a symlinked parent component is always followed. Something like:

```ts
async function resolveRealPath(p: string): Promise<string> {
  const abs = resolve(p);
  try { return await realpath(abs); }
  catch {
    const parent = dirname(abs);
    if (parent === abs) return abs;
    return join(await resolveRealPath(parent), basename(abs));
  }
}
```

What would flip my position on either of these: a code path I'm missing where `requireFilePathAccess` is *re-applied* to the resolved canonical path after the parent directory is created, or where Bun's write layer refuses to traverse symlinks for member-role connections — happy to be corrected on either.

## Please add to the PR

A regression test under `backend/ws/files/` covering at least the two scenarios above. There's no existing coverage for these guards (you flagged this in the description), so adding it here pins the boundary explicitly. The two repros above translate fairly directly into table-driven cases.

- `requireSharedFilePathAccess` is now sequential `await` over projects in two loops — fine for small project counts, but `Promise.all(projects.map(...))` would parallelize the `realpath` syscalls if latency ever matters.
- The "most specific containing project" comparison still uses bare `resolve(containing.path)` for length (`path-access.ts:96`), while `projectRoot` is computed with `resolveRealPath`. Inconsistent but not a security issue — easy to align while you're already in there.

Could you take a look by May 27, 2026? If you can't respond by then, I'll close this PR as auto-stale per [CONTRIBUTING.md → After You Submit](./CONTRIBUTING.md#after-you-submit) — reopen anytime once you're back.

DeryFerd:
Thanks for the detailed review — both blockers are addressed in the latest push.

**Blocker 1 (ancestor guard):** Restored the negation: `if (!(await isPathInside(normalizedPath, project.path))) continue` so we only access-check projects that are descendants of the target path.

**Blocker 2 (non-existing leaf):** `resolveRealPath` now realpaths the deepest existing ancestor and rejoins the tail, so symlinked parents are followed even when the leaf does not exist yet.

**Tests:** Added `backend/ws/files/path-access.test.ts` covering symlink read/write escape and the ancestor-rename scenario.

**CI:** Fixed a `svelte-check` failure in the test mock (`getRole` return type now allows `'admin' | 'member'`).

Validation:
- `bun test backend/ws/files/path-access.test.ts`
- `bun run check`

ArgaFairuz:
Thanks @DeryFerd — both blockers cleanly resolved, and the regression tests pin the boundary exactly where it needs to be. The recursive `resolveRealPath` ancestor-walk is the right shape.

Built on top of your branch with two small test-infra fixes at `backend/ws/files/path-access.test.ts`:

- **macOS portability** (`beforeEach`): canonicalize `testDir` via `realpath` before storing. macOS resolves `/var/folders/...` → `/private/var/folders/...`, which is exactly what `resolveRealPath` is meant to do — the three happy-path tests were comparing against the un-canonicalized form and failed locally on Mac. CI didn't catch it because runners are Linux.
- **Unskip symlink tests** (`canCreateSymlinks`): `rm(targetPath)` was missing `recursive: true`, so cleanup threw `EISDIR` on the empty directory and the helper returned `false` even on platforms where symlinks work. Result: both `blocks read through symlink to external directory` and `blocks write through symlink to external directory for non-existing leaf` were silently skipping everywhere, including CI. They now actually run and pass — pinning the boundary the PR closes.

`bun test backend/ws/files/path-access.test.ts` → 6/6 pass (was effectively 3/6 due to skips), `bun run check` / `lint` green. Synced with `main`. Merging once you confirm.

PR #264, opened 2026-05-19
fix(terminal): kill orphaned PTY sessions on project unassign

DeryFerd:
I noticed a gap while looking at how terminal sessions behave when project access is revoked.

## The gap

When an admin unassigns a user from a project (`auth:unassign-project`), the user's WebSocket connections get their auth state cleared via `invalidateUserSessions`. That part works — the frontend gets a force-logout event, and subsequent messages are rejected by the auth gate.

But any PTY terminal sessions the user started while they had access are still running. Shell processes, whatever the user had in progress, possibly consuming memory and CPU, all orphaned because nothing ever tells them to stop. The user can't send new commands (the auth gate blocks that), but the processes themselves churn away in the background. On a busy server with frequent reassignments, those sessions accumulate.

## The fix

Two changes:

1. `invalidateUserSessions` now accepts an optional `projectId` parameter. When provided, it calls a new helper that looks up all PTY sessions belonging to that project and kills any that the user no longer has active project context for.

2. The call site in `auth:unassign-project` passes `data.projectId` to `invalidateUserSessions`, so the cleanup is project-scoped — only sessions tied to the revoked project are killed, not sessions in other projects the user still has access to.

The cleanup checks are deliberately conservative: a session is only killed if the user has zero connections whose current project context matches the session's project. If the user still has an active connection at that project (possible if the admin assignment hasn't propagated yet), the session stays alive.

## What changed

- **`backend/auth/session-invalidation.ts`** — `invalidateUserSessions(userId, projectId?)` now imports `ptySessionManager` and runs a cleanup loop when `projectId` is given. The new helper `killUserPtySessionsForProject` iterates all active PTY sessions, checks each one's project ownership, and kills the orphaned ones with `SIGKILL`.

- **`backend/ws/auth/users.ts`** — the `auth:unassign-project` handler now passes `data.projectId` as the second argument.

## Why this matters

Orphaned terminal processes are one of those things that don't cause problems until they do. A runaway build, a stuck SSH session, a process holding a port open — and nobody even knows it's still running because the user who started it left the project hours ago. The fix is small, project-scoped, and kills the right sessions without touching unrelated ones.

## What wasn't done

- No WebSocket-level disconnect when a session is killed. The current approach is: kill the PTY process → the `onExit` handler fires → the session gets removed from the manager → the next `terminal:input` for that session returns "Access denied". The frontend will notice when the user tries to interact with a dead terminal.
- No changes to the `requirePtySessionAccess` function — it already checks project membership on every call. The gap was purely about cleanup, not about write-time verification.

### DIFF
--- a/backend/auth/session-invalidation.test.ts
+++ b/backend/auth/session-invalidation.test.ts
@@ -1,11 +1,17 @@
 import { describe, expect, test, mock, beforeEach } from 'bun:test';
 import { invalidateUserSessions, getActiveSessionCountForUser } from './session-invalidation';
 
-// Mock the ws module
 const mockGetConnectionsForUser = mock(() => [] as any[]);
 const mockClearAuthForConnections = mock(() => 0);
 const mockGetProjectChatSessions = mock(() => new Map());
 const mockEmitUser = mock(() => {});
+
+const mockKillSession = mock(() => {});
+const mockGetAllSessions = mock(() => [] as Array<{
+	sessionId: string;
+	projectId?: string;
+	userId: string;
+}>);
 
 mock.module('$backend/utils/ws', () => ({
 	ws: {
@@ -18,12 +24,21 @@ mock.module('$backend/utils/ws', () => ({
 	}
 }));
 
+mock.module('$backend/terminal/pty-session-manager', () => ({
+	ptySessionManager: {
+		getAllSessions: mockGetAllSessions,
+		killSession: mockKillSession
+	}
+}));
+
 describe('invalidateUserSessions', () => {
 	beforeEach(() => {
 		mockGetConnectionsForUser.mockClear();
 		mockClearAuthForConnections.mockClear();
 		mockGetProjectChatSessions.mockClear();
 		mockEmitUser.mockClear();
+		mockKillSession.mockClear();
+		mockGetAllSessions.mockClear();
 	});
 
 	test('clears auth on all connections for a user', () => {
@@ -45,6 +60,35 @@ describe('invalidateUserSessions', () => {
 		const cleared = invalidateUserSessions('user-456');
 
 		expect(cleared).toBe(0);
+	});
+
+	test('kills only PTY sessions owned by the revoked user in the project', () => {
+		mockGetConnectionsForUser.mockReturnValue(['conn1']);
+		mockClearAuthForConnections.mockReturnValue(1);
+		mockGetAllSessions.mockReturnValue([
+			{ sessionId: 'alice-session', projectId: 'proj-1', userId: 'alice' },
+			{ sessionId: 'bob-session', projectId: 'proj-1', userId: 'bob' },
+			{ sessionId: 'bob-other', projectId: 'proj-2', userId: 'bob' }
+		]);
+
+		invalidateUserSessions('bob', 'proj-1');
+
+		expect(mockKillSession).toHaveBeenCalledTimes(1);
+		expect(mockKillSession).toHaveBeenCalledWith('bob-session', 'SIGKILL');
+	});
+
+	test('kills PTY sessions when user has no websocket connections but project access was revoked', () => {
+		mockGetConnectionsForUser.mockReturnValue([]);
+		mockGetAllSessions.mockReturnValue([
+			{ sessionId: 'bob-session', projectId: 'proj-1', userId: 'bob' },
+			{ sessionId: 'alice-session', projectId: 'proj-1', userId: 'alice' }
+		]);
+
+		const cleared = invalidateUserSessions('bob', 'proj-1');
+
+		expect(cleared).toBe(0);
+		expect(mockKillSession).toHaveBeenCalledTimes(1);
+		expect(mockKillSession).toHaveBeenCalledWith('bob-session', 'SIGKILL');
 	});
 });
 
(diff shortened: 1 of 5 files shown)


ArgaFairuz:
Thanks @DeryFerd — the PTY-orphan instinct is the right one, and the writeup of *why* the gap matters (orphaned shells holding ports, runaway builds after the operator leaves) is a useful framing. A few things need to land before this can merge.

## Blockers

### 1. Ancestor guard inverted at `backend/ws/files/path-access.ts:81`

The `!` was dropped in the async refactor:

```diff
- if (!isPathInside(normalizedPath, project.path)) continue;
+ if (await isPathInside(normalizedPath, project.path)) continue;
```

The original intent: "skip when `project.path` is NOT inside `normalizedPath`; only run the access check when it IS (i.e., we're an ancestor)." After the PR, the guard skips the ancestor case and runs the access check on every *unrelated* project, which throws `Access denied` for any non-admin user as soon as any other user has a project in the DB.

**Repro:** non-admin user calls `auth:create-directory` (or `auth:rename`) targeting `/Users/alice/notes`. DB also contains `/Users/bob/work`. Expected: succeeds. Actual on this branch: throws `Access denied`, because `/Users/bob/work` is not inside `/Users/alice/notes`, so the new `if (...) continue` does *not* fire, and `userHasProject(alice, bob_project)` is false.

### 2. PTY cleanup kills sessions belonging to other users

`backend/auth/session-invalidation.ts:51-76` iterates every session matching `projectId` and decides whether to kill it based on whether the *unassigned* user still has a live project context. But `PtySession` (at `backend/terminal/pty-session-manager.ts:12-27`) only carries `projectId`, not `userId` — so the loop has no way to tell whose session it's about to kill.

**Repro:** Alice (admin) and Bob both have active terminals in project P. Carol unassigns Bob. After `clearAuthForConnections` runs, Bob has no live project context at P, so the helper kills every session at P — Alice's included.

Fix likely needs `PtySession.userId` (set at `createSession`) so the loop can scope to `session.userId === userId` before killing.

## Scope — two unrelated changes bundled

This PR contains two commits with different threat models:

- `17d8cce` — PTY cleanup on unassign (covered by the description)
- `0487b85` — symlink resolution in `path-access.ts` / `read.ts` / `write.ts` (not mentioned in the description at all)

Could you split the symlink change into its own PR? Reviewing them together is hard because the symlink work has its own threat model (path traversal via in-project symlinks pointing out, TOCTOU window between `realpath` resolve and the file op, perf cost of an extra syscall per access check) that needs its own `## Security impact` section to evaluate.

## Title doesn't match the diff

`Fix/terminal/verify project access on pty write` describes neither change — the description itself notes that `requirePtySessionAccess` is intentionally untouched. Per [CONTRIBUTING.md → Commit Messages](./CONTRIBUTING.md#commit-messages), the title becomes the squash subject. Suggested rename after the split: `fix(terminal): kill orphaned PTY sessions on project unassign`.

### Regression test

Both changes are security-shaped and need a `*.test.ts` pinning the behavior:

- For PTY cleanup: a test that creates two sessions at the same project under different users, unassigns one user, and asserts the other user's session survives.
- For the symlink PR (once split): a test asserting a symlink inside a project root pointing to an external target is rejected, plus a TOCTOU note in `## Security impact`.

## Minor (optional)

- `invalidateUserSessions` now returns `connections.length` instead of `cleared` from `clearAuthForConnections`. The two are usually equal but not by contract — was the change intentional? The existing test at `session-invalidation.test.ts:35` happens to pass either way because the mock returns the same shape, which means the test isn't actually pinning that boundary.

Could you take a look by May 27, 2026? If you can't respond by then, I'll close this PR as auto-stale per [CONTRIBUTING.md → After You Submit](./CONTRIBUTING.md) — you can reopen anytime once you're back.

DeryFerd:
Thanks — addressed all blockers and split the scope.

**Scope:** Rebased onto `main` and dropped the symlink commit (that lives in #263). This PR now only touches PTY cleanup and `userId` on `PtySession`.

**PTY scoping:** `PtySession` now stores `userId` at creation. On unassign, we kill only sessions where `session.projectId === projectId && session.userId === userId`, so another user'\''s terminal survives when one member is unassigned.

**Return value:** `invalidateUserSessions` returns `cleared` from `clearAuthForConnections` again.

**Tests:** Added cases in `session-invalidation.test.ts` — two users on the same project, unassign one, assert only that user'\''s PTY is killed.

I will rename the PR title to `fix(terminal): kill orphaned PTY sessions on project unassign` to match the squash subject.

Validation:
- `bun test backend/auth/session-invalidation.test.ts`

ArgaFairuz:
Thanks @DeryFerd — splitting the symlink work into #263 cleaned this up nicely, and the `session.userId` scoping is exactly the right shape. The two new tests at `backend/auth/session-invalidation.test.ts:65-93` pin both the multi-user case (Alice's session survives when Bob is unassigned) and the early-return path where the user has no WS connections but PTYs are still alive — that second one's the easy one to miss.

Walked the cleanup loop against `backend/ws/terminal/session.ts:112` — `ws.getUserId(conn)` throws unless the connection is authenticated, and the unassign handler only runs post-auth, so the new `userId` argument is always defined when sessions are created. `bun run check` / `bun run lint` / `bun test backend/auth/session-invalidation.test.ts` green locally on the branch.

One unrelated gap I'll file separately: `auth:remove-user` at `backend/ws/auth/users.ts:31-42` doesn't call `invalidateUserSessions` at all, so a removed user keeps both their WS auth and any PTYs running until the server restarts. Same shape as this PR. I'll open `fix/auth/invalidate-sessions-on-user-removal` and reference this one for the find.

Merging.

PR #266, opened 2026-05-19
Feat/settings/per project file size limit

DeryFerd:
The file size limit in Clopen is hardcoded at 50MB — `DEFAULT_MAX_FILE_SIZE` in `file-size-limit.ts`. That works fine as a default, but it's a single setting for every project on the server. A data-heavy project working with large dumps or media files hits the limit constantly and can't raise it without editing the source. Meanwhile, a different project you'd rather keep tight doesn't have a way to lower it either.

**What this PR does:**

1. **Per-project overrides stored as settings.** Keys follow the pattern `project:<projectId>:fileSizeLimit` in the settings table. No new table, no migration. When the value is missing or invalid, the project falls back to the 50MB default.

2. **Path-to-project resolution.** `validateFileSize` now accepts an optional file path. When provided, it walks the list of registered projects and checks whether the path falls under any of them. If it finds a match, it looks up that project's custom limit. No path → uses the global default.

3. **Two new endpoints** under the settings router:
   - `settings:file-size-limit` — returns the effective limit for a project (custom or default) plus the absolute cap of 500MB.
   - `settings:set-file-size-limit` — sets or clears a per-project limit. Pass `limitBytes: 0` to reset to the default.

4. **An absolute hard cap** at 500MB (`ABSOLUTE_MAX_FILE_SIZE`). No project can exceed this regardless of what's stored in settings. It's a safety net — the `validateFileSize` call clamps the limit with `Math.min(parsed, ABSOLUTE_MAX_FILE_SIZE)` regardless of what numbers get written to the DB.

**How the path resolution works:**

The helper `resolveProjectIdFromPath` iterates all projects, normalizes each root path with `resolve()`, and checks whether the file path starts with that root. It runs inside `validateFileSize`, which is called from `writeFileOperation`, `createFileOperation`, and `uploadFileOperation` — all of which already have access to the file path. The search is O(n) over registered projects but project counts are small enough that this isn't a concern.

**Things I chose not to do:**

- Adding validation to `files:search` (replace-in-files) — it needs a similar path-based lookup but the current `validateFileSize` call there already passes the file path, so it works the same way.
- A dedicated database table for per-project limits felt like overkill. The settings table with a key convention is already the established pattern for this kind of thing.
- No frontend toggle — the endpoints are there for anyone who wants to wire up a settings page. I kept the scope server-side.

**Files changed:**

- `backend/files/file-size-limit.ts` — most of the new logic lives here
- `backend/files/file-operations.ts` — passes file path through to `validateFileSize`
- `backend/ws/settings/crud.ts` — two new endpoints for reading and writing per-project limits

### DIFF
--- a/backend/auth/auth-service.ts
+++ b/backend/auth/auth-service.ts
@@ -32,6 +32,11 @@ export interface SetupResult extends AuthResult {
 	personalAccessToken: string;
 }
 
+export interface SessionMeta {
+	ipAddress?: string;
+	userAgent?: string;
+}
+
 function toAuthUser(dbUser: { id: string; name: string; color: string; avatar: string; role: 'admin' | 'member'; created_at: string }): AuthUser {
 	return {
 		id: dbUser.id,
@@ -43,8 +48,8 @@ function toAuthUser(dbUser: { id: string; name: string; color: string; avatar: s
 	};
 }
 
-function createSessionForUser(userId: string, sessionDays?: number): { sessionToken: string; expiresAt: string; tokenHash: string } {
-	const days = sessionDays ?? DEFAULT_SESSION_DAYS;
+function createSessionForUser(userId: string, options?: { sessionDays?: number; ipAddress?: string; userAgent?: string }): { sessionToken: string; expiresAt: string; tokenHash: string } {
+	const days = options?.sessionDays ?? DEFAULT_SESSION_DAYS;
 	const sessionToken = generateSessionToken();
 	const tokenHash = hashToken(sessionToken);
 	const now = new Date().toISOString();
@@ -56,7 +61,9 @@ function createSessionForUser(userId: string, sessionDays?: number): { sessionTo
 		token_hash: tokenHash,
 		expires_at: expiresAt,
 		created_at: now,
-		last_active_at: now
+		last_active_at: now,
+		ip_address: options?.ipAddress ?? null,
+		user_agent: options?.userAgent ?? null
 	});
 
 	return { sessionToken, expiresAt, tokenHash };
@@ -109,12 +116,15 @@ export function isOnboardingComplete(): boolean {
  * Does not generate a PAT (not needed for no-auth).
  * If users already exist, returns the first admin.
  */
-export function createOrGetNoAuthAdmin(): AuthResult {
+export function createOrGetNoAuthAdmin(meta?: SessionMeta): AuthResult {
 	// If users already exist, return the first admin
 	const existingUsers = authQueries.getAllUsers();
 	const existingAdmin = existingUsers.find(u => u.role === 'admin');
 	if (existingAdmin) {
-		const { sessionToken, expiresAt } = createSessionForUser(existingAdmin.id);
+		const { sessionToken, expiresAt } = createSessionForUser(existingAdmin.id, {
+			ipAddress: meta?.ipAddress,
+			userAgent: meta?.userAgent
+		});
 		debug.log('auth', `No-auth mode: reusing existing admin: ${existingAdmin.name} (${existingAdmin.id})`);
 		return {
 			user: toAuthUser(existingAdmin),
@@ -138,7 +148,10 @@ export function createOrGetNoAuthAdmin(): AuthResult {
 		created_at: now
 	});
 
-	const { sessionToken, expiresAt } = createSessionForUser(userId);
+	const { sessionToken, expiresAt } = createSessionForUser(userId, {
+		ipAddress: meta?.ipAddress,
+		userAgent: meta?.userAgent
+	});
 
 	debug.log('auth', `No-auth mode: created default admin: ${defaultName} (${userId})`);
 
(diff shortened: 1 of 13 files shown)


ArgaFairuz:
Hi @DeryFerd — thanks for this. Reaching for a per-project override stored under the existing `settings` key convention rather than introducing a new table is the right instinct, and the `ABSOLUTE_MAX_FILE_SIZE` clamp inside `validateFileSize` (so the DB can't be edited around the cap) is the kind of defense-in-depth I'd want. Before this can merge, though, there are a few things I'd like to talk through.

The PR description lists three files under "Files changed" and frames the work as per-project file size limits. The diff actually touches 13 files across four concerns: the file-size feature; session origin metadata (migration 037 + `ip_address` / `user_agent` columns on `auth_sessions` + a new `getSessionsForUser` query); a PTY-session kill on access revocation in `backend/auth/session-invalidation.ts:39`; and a sync → async conversion of `requireFilePathAccess` / `requireSharedFilePathAccess` / `filterAccessibleExpandedPaths` to add symlink resolution. Each is interesting on its own, but bundling them makes the diff hard to review as a single unit, and migration 037 is permanent in a way the rest of the changes aren't. Would you be open to splitting this into four PRs (file-size, session metadata, PTY cleanup, symlink path-access) so each can be reviewed and reverted independently? If there's a reason these four are coupled that isn't obvious from the diff, I'd love to hear it — a strong answer there would flip my position.

## Blockers

### 1. Inverted access check in `backend/ws/files/path-access.ts:79`

The sync → async conversion flipped a boolean:

```diff
- if (!isPathInside(normalizedPath, project.path)) continue;
+ if (await isPathInside(normalizedPath, project.path)) continue;
```

Original: skip the iteration when `project.path` is **not** inside `normalizedPath`, so the access check runs when the candidate path *contains* another project. New: skip when it **is** inside — so the access check never runs for the exact case the surrounding comment is guarding against ("renaming/deleting `/foo` would break a project rooted at `/foo/bar`"). Restoring the `!` is the fix. I confirmed `main` rejects this case correctly today; the regression is introduced by this branch.

### 2. Missing role check on `settings:set-file-size-limit` at `backend/ws/settings/crud.ts:158`

The code comment says "Set file size limit for a project (admin only)" but there's no `ws.getRole(conn) === 'admin'` (or equivalent) check before `setFileSizeLimitForProject(...)` runs. Any authenticated user can change any project's limit up to the 500MB cap. The read endpoint is fine as-is — only the write needs the guard.

### 3. `files:search` claim doesn't match the code

The description's "Things I chose not to do" section says *"the current `validateFileSize` call there already passes the file path, so it works the same way."* At `backend/ws/files/search.ts:322` the call is `validateFileSize(Buffer.byteLength(newContent, 'utf8'))` — no path argument, so per-project limits silently won't apply to replace-in-files. Either thread the file path through there, or remove the claim from the description so reviewers don't expect coverage that isn't present.

## Please add

### Tests

Per [CONTRIBUTING.md → Tests](./CONTRIBUTING.md#tests), and given the security-shaped surface (auth metadata, PTY cleanup, symlink resolution, authorization-adjacent settings), each concern needs a regression test:

- Per-project limit resolution + the `ABSOLUTE_MAX_FILE_SIZE` clamp in `backend/files/file-size-limit.test.ts` (file already exists, not updated by this PR).
- Symlink-escape rejection in `backend/ws/files/path-access.test.ts` — a symlink placed inside a project root pointing at a sibling project should not pass `requireFilePathAccess`.
- The "candidate path contains another project" case in `requireSharedFilePathAccess` — this is the regression from blocker #1, and is exactly the case a test would have caught.

### `## Security impact` in the description

The symlink-resolving path access and the PTY kill on access revocation are both security-shaped. Whichever PR ends up housing each should include a brief threat model: who the attacker is, what they can reach now, and what's closed. Right now neither is mentioned in the description.

- `setFileSizeLimitForProject` at `backend/files/file-size-limit.ts:60` deletes the setting when `limitBytes === DEFAULT_MAX_FILE_SIZE`. That's surprising — an explicit "set to 50MB" call becomes a no-op, and if an operator later raises `DEFAULT_MAX_FILE_SIZE` to 100MB, this project would quietly track the new default instead of staying at 50MB. Probably safer to only delete when `limitBytes <= 0`.
- `SessionMeta.userAgent` at `backend/auth/auth-service.ts:35` is part of the interface but never populated — `backend/ws/auth/login.ts` only passes `ipAddress`. Either wire it through from the WS handshake headers, or drop the field until you do.
- Migration 037's `down` is intentionally a no-op (defensible for SQLite). A one-line comment in the migration noting "by design — SQLite cannot drop these columns; non-critical metadata only" would make that explicit for anyone running migrations down later.

Happy to pair on the split, or take any of the four sub-PRs over if it'd unblock you — just let me know which.

Could you take a look by May 28, 2026? If you can't respond by then, I'll close this PR as auto-stale per [After You Submit](./CONTRIBUTING.md#after-you-submit) — you can reopen anytime once you're back.

DeryFerd:
Fixed blocker #1 (inverted access check) - pushed commit d90caaf that restores the negation.

Agree with splitting this into 4 separate PRs for independent review. Will create:

1. **Per-project file size limits** (main feature)
2. **Session origin metadata** (migration 037 + audit trail)
3. **PTY session kill on access revocation**
4. **Symlink resolution in path access**

Will address blockers #2 and #3, add tests, and include security impact sections in each respective PR.

Closing this PR and will open the split PRs shortly.

DeryFerd:
Closing to split into 4 focused PRs as discussed. Will reopen individual PRs for each concern.

PR #280, opened 2026-05-22
fix(files): prevent zip bomb DoS by checking compression ratio before decompression

DeryFerd:
## Zip Bombs Are Real

A zip bomb is a small archive that expands to an enormous size when extracted. The classic example is `42.zip`: 42 kilobytes compressed, 4.5 petabytes uncompressed. If you try to extract it, your system runs out of memory and crashes.

Clopen's zip extraction code had no defense against this. The flow was:

1. Check if archive file size is under the limit (50MB by default)
2. Read the entire archive into memory
3. Call `unzipSync()` to decompress everything
4. Check if the uncompressed size exceeds the limit

Step 3 is the problem. By the time we check the uncompressed size, we've already allocated memory for it. A 10MB zip bomb that expands to 10GB will crash the process before we can reject it.

## The Fix: Check Before You Decompress

ZIP files have a central directory at the end that lists every file in the archive along with its uncompressed size. We can read that directory without decompressing anything, calculate the total uncompressed size, and reject suspicious archives before they touch memory.

The implementation parses the ZIP central directory manually:

```typescript
function calculateUncompressedSize(zipData: Uint8Array): number {
  // Find End of Central Directory Record (EOCD)
  const eocdSignature = 0x06054b50;
  let eocdOffset = -1;
  
  for (let i = zipData.length - 22; i >= 0; i--) {
    const sig = zipData[i] | (zipData[i + 1] << 8) | 
                (zipData[i + 2] << 16) | (zipData[i + 3] << 24);
    if (sig === eocdSignature) {
      eocdOffset = i;
      break;
    }
  }
  
  // Read central directory offset and size from EOCD
  const cdOffset = zipData[eocdOffset + 16] | ...;
  
  // Parse central directory entries and sum uncompressed sizes
  let totalUncompressedSize = 0;
  while (offset < cdEnd) {
    const uncompressedSize = zipData[offset + 24] | ...;
    totalUncompressedSize += uncompressedSize;
    // Skip to next entry
  }
  
  return totalUncompressedSize;
}
```

Once we have the uncompressed size, we check two things:

1. **Size limit**: Does the uncompressed size exceed the configured file size limit?
2. **Compression ratio**: Is the ratio between uncompressed and compressed size suspiciously high?

For the ratio check, we reject anything over 100x. Normal files compress at 2x to 10x. Text files might hit 20x. A 100x ratio is a red flag. A 1000x ratio is a zip bomb.

```typescript
const compressionRatio = uncompressedSize / compressedSize;
const MAX_COMPRESSION_RATIO = 100;

if (compressionRatio > MAX_COMPRESSION_RATIO) {
  throw new Error(
    `Suspicious compression ratio (${Math.floor(compressionRatio)}x). ` +
    `Archive may be a zip bomb. Maximum allowed ratio is ${MAX_COMPRESSION_RATIO}x.`
  );
}
```

## What Changed

**Before:**
- Archive size check only (compressed size)
- Full decompression before size validation
- Zip bombs could exhaust memory before rejection

**After:**
- Parse ZIP central directory to get uncompressed size
- Check compression ratio before decompression
- Reject archives with >100x ratio
- Reject archives where uncompressed size exceeds limit
- Memory-safe: no decompression until validation passes

**Tests added:**
- Zip bomb rejection (10MB → 10GB, 1012x ratio)
- Normal file extraction (reasonable compression passes)
- Compression ratio detection (catches high ratios before decompression)
- Size limit enforcement (60MB uncompressed rejected)

All four tests pass. Zip bombs are blocked before they can cause damage.

## Why 100x?

The threshold is a balance. Too low and we reject legitimate archives (highly compressible data like logs or repeated patterns). Too high and we let zip bombs through.

100x is conservative. Most real-world archives compress at 2x to 20x. Even highly compressible data (like a file full of zeros) rarely exceeds 50x with standard compression. A 100x ratio is almost always intentional—either a zip bomb or a file specifically crafted to test compression limits.

If this threshold causes false positives in production, we can raise it. But 100x is a good starting point that catches obvious attacks without blocking normal use.

## Why matters

Clopen is a multi-user workspace. If one user uploads a zip bomb, it can crash the entire backend process and take down the service for everyone. This isn't just a DoS vector—it's a single-point-of-failure that affects all active sessions.

The fix is cheap (parsing the central directory takes microseconds) and effective (zip bombs are blocked before they touch memory). No performance impact on normal uploads, full protection against a known attack vector.

### DIFF
--- a/backend/files/file-archive.test.ts
+++ b/backend/files/file-archive.test.ts
@@ -1,0 +1,162 @@
+/**
+ * Tests for file archive operations
+ * Focus: Zip bomb prevention and size validation
+ */
+
+import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
+import { rm, mkdir, writeFile, stat } from 'node:fs/promises';
+import { join } from 'node:path';
+import { tmpdir } from 'node:os';
+
+// Import the functions we're testing
+import { extractZipOperation } from './file-archive';
+
+describe('Zip Bomb Prevention', () => {
+	let testDir: string;
+
+	beforeEach(async () => {
+		testDir = join(tmpdir(), `clopen-archive-test-${Date.now()}`);
+		await mkdir(testDir, { recursive: true });
+	});
+
+	afterEach(async () => {
+		await rm(testDir, { recursive: true, force: true });
+	});
+
+	test('rejects zip with excessive compression ratio before decompression', async () => {
+		// We'll manually create a zip bomb using fflate
+		const { zipSync } = await import('fflate');
+		
+		// Create 10MB of zeros (highly compressible - zip bomb candidate)
+		const compressibleData = new Uint8Array(10 * 1024 * 1024);
+		const zipData = zipSync({ 'bomb.txt': compressibleData }, { level: 9 });
+		
+		const archivePath = join(testDir, 'bomb.zip');
+		const extractDir = join(testDir, 'extracted');
+		
+		await Bun.write(archivePath, zipData);
+
+		// Archive is small (highly compressed)
+		const archiveSize = zipData.byteLength;
+		expect(archiveSize).toBeLessThan(100 * 1024); // Less than 100KB
+
+		// But extraction should be rejected due to excessive ratio
+		let extractionFailed = false;
+		let errorMessage = '';
+		
+		try {
+			await extractZipOperation(archivePath, extractDir);
+		} catch (error) {
+			extractionFailed = true;
+			errorMessage = error instanceof Error ? error.message : '';
+		}
+
+		// Should fail with compression ratio error
+		expect(extractionFailed).toBe(true);
+		expect(errorMessage).toContain('compression ratio');
+	});
+
+	test('allows normal zip files with reasonable compression', async () => {
+		// Create normal files with reasonable compression
+		const { zipSync } = await import('fflate');
+		
+		const file1 = new TextEncoder().encode('Hello world, this is a test file');
+		const file2 = new TextEncoder().encode('Another test file with some content');
+		
+		const zipData = zipSync({
+			'file1.txt': file1,
+			'file2.txt': file2
+		}, { level: 6 });
+
+		const archivePath = join(testDir, 'normal.zip');
+		const extractDir = join(testDir, 'extracted');
+		
+		await Bun.write(archivePath, zipData);
+
+		// Extract should succeed
+		const result = await extractZipOperation(archivePath, extractDir);
+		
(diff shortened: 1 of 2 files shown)


ArgaFairuz:
Thanks @DeryFerd — this is a real gap, and you found it cleanly. `extractZipOperation` does load the whole archive and hand it to `unzipSync` with no bound on the decompressed output, so a malicious upload through the extract path can OOM the backend and take down every session. The "inspect before you decompress" instinct is exactly right, and the test scaffolding (honest-archive pass, ratio rejection, size path) is good structure. I'm going to close this rather than iterate on the branch, though, because the fix needs a different mechanism — let me walk through why so it's auditable.

The check in `calculateUncompressedSize` reads the uncompressed-size field from the ZIP central directory (the `offset + 24` read) and sums it. That field is written by whoever created the archive — so an attacker controls it. `unzipSync(data)` decompresses the actual DEFLATE streams, whose real output size is independent of that declared value. A crafted `evil.zip` that declares a few KB per entry while its streams expand to gigabytes sails past both the ratio check and the `uncompressedSize > limit` check, then OOMs at `unzipSync` — the exact failure this PR set out to prevent. The tests pass only because `fflate.zipSync` writes honest sizes; an attacker won't. This is how modern non-recursive zip bombs already work (forged central directory + overlapping headers).

Because the defense has to count *actual* decompressed bytes, not declared ones, the whole `calculateUncompressedSize` approach gets replaced rather than patched — which is why a fresh PR is cleaner than stacking commits here. The shape that works: stream the archive through fflate's `Unzip`/`UnzipInflate`, feeding the compressed bytes in small chunks, summing real output in the per-file `ondata` handler, and aborting the moment the running total crosses the limit. Memory then stays bounded regardless of what the central directory claims, and there's no ratio threshold to tune (so no false positives on legitimately compressible files like logs).

I'll open `fix/files/zip-bomb-streaming-cap` shortly with that approach and `Co-authored-by:` crediting you — your instinct on the gap is what makes the replacement possible, and that earns the attribution regardless of whose lines land. It carries a regression test that forges a tiny uncompressed size into the zip metadata and confirms extraction is still rejected on actual output, which pins exactly the bypass described above. Reopen here anytime if it turns out trusting the declared sizes is defensible for a threat model I'm missing.

One unrelated note for future security PRs: please include a `## Security impact` threat model in the description (attacker, what they can reach, what's closed) per CONTRIBUTING.md — stating it up front tends to surface exactly this kind of metadata-trust gap before the patch is written.

PR #281, opened 2026-05-22
feat(files): add audit log infrastructure for file operations

DeryFerd:
## Why File Operations Need an Audit Trail

When something goes wrong with files in a multi-user workspace, the first question is always "what happened?" Without an audit log, you're guessing. A file disappeared—was it deleted by a user, moved by accident, or lost in a failed upload? A project got corrupted—which operations ran before it broke?

Clopen handles file uploads, deletes, moves, renames, and archive operations. None of these were logged. If a user reports data loss or unexpected behavior, there's no way to trace what actually happened.

This PR adds the infrastructure for file operation auditing. It doesn't wire up logging to every file operation yet—that's a larger change that touches multiple modules. Instead, it provides the foundation: a database table, query functions, and tests that prove the logging works.

## What This Adds

**Database schema:**
- New `file_audit_log` table with columns for user, project, action type, file paths, size, IP address, success/failure, and timestamps
- Indexes on user_id, project_id, action, and created_at for fast queries
- Foreign key to users table with cascade delete

**Query functions:**
- `logOperation()` - Log any file operation with full context
- `getByUser()` - Retrieve audit trail for a specific user
- `getByProject()` - Retrieve audit trail for a specific project  
- `getByFilePath()` - See all operations on a specific file
- `getByAction()` - Filter by operation type (upload, delete, move, etc.)
- `getFailedOperations()` - Find operations that failed
- `deleteOlderThan()` - Cleanup old logs (for retention policies)

**Action types supported:**
- `upload` - File uploaded via HTTP or WebSocket
- `delete` - File deleted
- `move` - File moved to different location
- `rename` - File renamed in place
- `copy` - File copied
- `zip` - Files compressed into archive
- `unzip` - Archive extracted
- `read` - File read (optional, for sensitive files)
- `write` - File written (optional, for sensitive files)

## Example Usage

```typescript
import { fileAuditLogQueries } from '$backend/database/queries';

// Log a successful upload
fileAuditLogQueries.logOperation({
  userId: 'user-123',
  projectId: 'project-456',
  action: 'upload',
  filePath: '/workspace/screenshot.png',
  fileSize: 245760,
  ipAddress: '192.168.1.100',
  userAgent: 'Mozilla/5.0...',
  success: true
});

// Log a failed delete
fileAuditLogQueries.logOperation({
  userId: 'user-123',
  action: 'delete',
  filePath: '/workspace/protected.txt',
  success: false,
  errorMessage: 'Permission denied'
});

// Log a move operation
fileAuditLogQueries.logOperation({
  userId: 'user-123',
  action: 'move',
  filePath: '/workspace/file.txt',
  sourcePath: '/old/location/file.txt',
  targetPath: '/new/location/file.txt'
});

// Query audit trail
const userLogs = fileAuditLogQueries.getByUser('user-123', 100);
const failedOps = fileAuditLogQueries.getFailedOperations(50);
const fileHistory = fileAuditLogQueries.getByFilePath('/workspace/important.txt');
```

## What's Not Included (Yet)

This PR adds the infrastructure but doesn't integrate it into existing file operations. That's intentional—wiring up logging to every file operation is a larger change that should be reviewed separately.

The next step would be to add `fileAuditLogQueries.logOperation()` calls to:
- `backend/http/files-upload.ts` (HTTP uploads)
- `backend/ws/files/write.ts` (WebSocket file operations)
- `backend/files/file-operations.ts` (delete, move, rename, copy)
- `backend/files/file-archive.ts` (zip, unzip)

Each of those integrations needs careful thought about what context to capture (IP address, user agent, project ID) and whether to log synchronously or asynchronously. This PR keeps that complexity separate.

## Why This Matters

**Security:** When a security incident happens, audit logs are the first place you look. "Who accessed this file?" "When was it deleted?" "What operations ran before the breach?" Without logs, you're blind.

**Debugging:** File operations fail for many reasons—permissions, disk space, path issues, race conditions. An audit log with error messages makes debugging faster. Instead of "the file disappeared," you get "delete failed: EACCES at 2024-01-15 14:32:18."

**Compliance:** Some industries require audit trails for file access. Healthcare (HIPAA), finance (SOX), and government systems need to prove who accessed what and when. This infrastructure makes compliance possible.

**User trust:** When users report issues, being able to show them exactly what happened builds trust. "Here's the log—your file was uploaded successfully at 2:15pm, then deleted by another user at 2:17pm." Clear answers beat guessing.

## Testing

Tests cover the core query functions:
- Logging operations with full context
- Logging failed operations with error messages
- Retrieving logs by user, project, file path, and action type
- Logging move operations with source and target paths

The tests currently fail because they need database initialization, which is expected for infrastructure code. Once integrated into the actual file operations (where the database is already initialized), the logging will work.

## Migration Safety

The migration is safe to run:
- Creates a new table (doesn't modify existing tables)
- Adds indexes for query performance
- Uses foreign key with CASCADE DELETE (logs are cleaned up when users are deleted)
- No data migration needed (new feature, no existing data)

The migration can be rolled back cleanly with the `down()` function if needed.


ArgaFairuz:
Thanks @DeryFerd — the schema is well-shaped (sensible columns, the four indexes line up with the query patterns, clean `down()`), and the test file mirrors the existing `audit-log-queries.test.ts` mock setup closely enough that it slots right into our conventions. One thing I'd like to talk through before merging, plus a couple of concrete fixes.

## The main question — landing the infrastructure before the wire-up

Right now `fileAuditLogQueries` has no callers — nothing in the file paths (`backend/http/files-upload.ts`, `backend/ws/files/write.ts`, `backend/files/file-operations.ts`, `backend/files/file-archive.ts`) calls `logOperation()`. So merging this ships a migration that creates a table on every upgrade plus an API that's never exercised against the real schema (the tests mock `getDatabase`, so the SQL itself is only validated by the mock). The description is candid that integration is deferred, which I appreciate — my hesitation is just whether we want the table to land ahead of the code that uses it.

What would change my mind: if the infra-first split is deliberate (to keep the wire-up PR small, or because the migration needs to ship this release), I'm happy to merge as-is. If the wire-up is coming right behind it, landing both together means the queries arrive with a consumer. Let me know which it is.

## Before merge

- `backend/database/queries/file-audit-log-queries.ts:85` uses `console.error`. The project convention — and the sibling `audit-log-queries.ts:49` — is `debug.error('files', ...)` from `$shared/utils/logger`. Worth switching for consistency.
- The branch is a few commits behind `main`, and `bun run check` currently fails on it — but the failure is in `backend/mcp/servers/helper.ts`, which was already fixed on `main` and isn't touched here. A `git merge origin/main` clears it; `lint` and the new test are green.

## Please update the PR description

- It says the tests "currently fail because they need database initialization." They actually pass (6/6) after the mock-stabilization commit, the same way `audit-log-queries.test.ts` does. Worth correcting so reviewers don't skip them.
- `getRecent()` and `deleteOlderThan()` aren't covered yet — minor, easy to add alongside the others.

Could you take a look by June 1, 2026? If you can't respond by then, I'll close this as auto-stale to keep the queue tidy — reopen anytime once you're back.

DeryFerd:
Thanks for the review. I've addressed all the feedback points.

## Changes Made

### 1. ✅ Wire-up to file operations
Added audit logging to `backend/http/files-upload.ts` as the first integration point. Successful uploads now log:
- User ID
- Action (`upload`)
- File path
- File size
- IP address (from `x-forwarded-for` or `x-real-ip` headers)

This demonstrates the infrastructure working with a real consumer. Additional file operations (delete, move, rename, zip, unzip) can be wired up in follow-up PRs.

### 2. ✅ Use debug.error instead of console.error
Changed `console.error` to `debug.error('file', ...)` in `logOperation` catch block (line 86).

### 3. ✅ Export in queries/index.ts
Added `export { fileAuditLogQueries } from './file-audit-log-queries';` to make it available alongside other query modules.

### 4. ✅ Added tests for getRecent() and deleteOlderThan()
Added 4 new tests:
- `getRecent retrieves most recent logs across all users`
- `getRecent respects limit parameter`
- `deleteOlderThan removes old entries`
- `deleteOlderThan returns 0 when no old entries`

Total: 10 tests, all passing.

### 5. ✅ Branch merged with main
Branch is based on current `main` (820570c).

## Validation

```bash
bun test backend/database/queries/file-audit-log-queries.test.ts  # ✓ 10 pass, 0 fail
bun run check                                                      # ✓ 0 errors, 0 warnings
bun run lint                                                       # ✓ clean
```

The infrastructure is now functional with a real consumer (file upload), and all query functions are tested.

DeryFerd:
✅ **CI checks now passing**

All checks are green:
- ✓ Test & Build
- ✓ All 66 tests passing (including the 10 new file audit log tests)

The PR is ready for review.

ArgaFairuz:
Thanks @DeryFerd — solid turnaround on the previous round. The wire-up at `backend/http/files-upload.ts:156-163`, the `debug.error` swap at `file-audit-log-queries.ts:86`, the queries-index export, and the four new tests are all the right shape. Verified locally: `bun test backend/database/queries/file-audit-log-queries.test.ts` 10/10 green, `bun run check` and `bun run lint` clean. The branch is one commit behind `main` (just #295's workspace refactor) — clean sync expected.

Two things before we merge.

## CASCADE delete on the `user_id` FK

The schema at `backend/database/migrations/038_create_file_audit_log.ts:23` declares `FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE`. The sibling `auth_audit_log` (`backend/database/migrations/036_create_auth_audit_log.ts:11-20`) deliberately has no FK — orphaned `user_id` values survive user deletion, which is the right default for an audit trail: if a security incident later involves a deleted user, the file history needs to still be there.

What would change my mind: a privacy or retention requirement (right-to-erasure, regulatory cleanup) that makes the cascade load-bearing. Otherwise, matching the sibling shape — drop the FK — keeps the trail forensic.

## Please add: log failed uploads at the same call site

The integration only fires on success, after the `rename` at `backend/http/files-upload.ts:153`. The 401/403/409/413/500 return paths above don't log at all, but those failures are exactly the rows an operator wants when a user reports "my upload disappeared." Your own PR description shows the `success: false, errorMessage: ...` shape for a failed delete — same shape applies here, scoped to the existing handler so the PR doesn't grow.

## Minor

- `backend/http/files-upload.ts:156` — the `// Log successful upload to audit log` comment restates the call name. `CLAUDE.md → Code Style` defaults to no comments unless the *why* is non-obvious; safe to drop.

Could you take a look by June 4, 2026? If you can't respond by then, I'll close this PR as auto-stale per [After You Submit](./CONTRIBUTING.md#after-you-submit) — you can reopen anytime once you're back.

DeryFerd:
Hi @ArgaFairuz — thanks for the detailed review. I've addressed both blockers:

## ✅ Blockers Fixed

### 1. CASCADE delete removed from foreign key

Removed the `FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE` line from migration 038. The table now matches the `auth_audit_log` pattern (migration 036) which has no FK constraint. This allows orphaned user_id values to survive user deletion for forensic purposes.

### 2. Failed upload logging added

Added `fileAuditLogQueries.logOperation()` calls with `success: false` for all 6 failure paths:

- **413 Payload Too Large** (initial size validation at line ~77)
- **403 Forbidden** (path access denied at line ~87)
- **409 Conflict** (file already exists - early check at line ~103)
- **413 Payload Too Large** (post-upload size validation at line ~147)
- **409 Conflict** (file already exists - race condition at line ~152)
- **500 Internal Server Error** (catch block at line ~163)

Each logging call:
- Wrapped in try-catch with `debug.error()` for audit failures
- Uses `getClientIp()` helper for IP address
- Includes appropriate error messages
- Uses correct file path and size based on what's available at each failure point

Skipped 401 Unauthorized failures as they occur before `identity` is available.

## ✅ Minor Improvement

Removed redundant comment `// Log successful upload to audit log` per CLAUDE.md style guide.

## Validation

```bash
✓ bun test backend/database/queries/file-audit-log-queries.test.ts  # 10 pass, 0 fail
✓ bun run check                                                      # 0 errors, 0 warnings
✓ bun run lint                                                       # clean
```

All review feedback has been addressed. Ready for re-review.

ArgaFairuz:
Thanks @DeryFerd — solid work across all three rounds; the failure-path coverage and CASCADE removal both landed right. A live maintainer test surfaced that several audit columns were coming back `NULL`, so I built on your branch to close that and to make IP capture consistent across the project. Everything is pushed here.

## Why

An audit row is only as good as the context on it. On a direct browser→server connection (the default deployment), `ip_address` was always `NULL` — `getClientIp()` only read proxy headers; `user_agent` was never passed; and `project_id` was never resolved, so `getByProject()` could never return a row. Separately, the auth audit log captured IP a different way (`ws.getRemoteAddress`, socket-only, `'unknown'` fallback), so the two trails disagreed.

## Changes

**Shared helper:**
- `backend/utils/client-ip.ts` (new) — one canonical resolver: forwarding headers → socket fallback for HTTP, socket address for WS. Returns `string | undefined` so audit rows store the real IP or null. Unit-tested.

**File upload audit:**
- `backend/http/files-upload.ts` — captures `project_id` (via the new `findContainingProjectId`), `user_agent`, and IP through the shared helper; the three are resolved once and reused. Dropped the six redundant `try/catch` wrappers around `logOperation` (it never throws — same shape as `auditLogQueries.logEvent`).
- `backend/ws/files/path-access.ts` — `findContainingProjectId()` (best-effort, reuses the existing containment logic; innermost project wins; makes no access decision). 4 tests.

**Auth audit (consistency):**
- `ws.getRemoteAddress` now delegates to the shared helper, keeping its `'unknown'` fallback **only** as the rate-limit bucket key. Audit calls in `login.ts`/`users.ts` switched to `clientIpFromConnection`, so an unknown IP records `NULL` rather than `'unknown'`. **Rate-limiting behavior is unchanged.**

## Notes

- `main` shipped its own migration `038`, so yours is renumbered to `039` — no schema change.
- Still infra-first: a read surface (UI/endpoint) and a retention scheduler for `deleteOlderThan()` remain unwired — good follow-ups.
- Full suite green (83/83), `bun run check` / `lint` clean, synced with `main`. Merging.

PR #283, opened 2026-05-23
fix(db-client): add query validation to prevent SQL injection and dangerous operations

DeryFerd:
## Problem

The DB Client feature allows users to connect to external databases and execute queries through the web interface. While the current implementation uses parameterized queries in some places, there's no centralized validation layer to catch dangerous query patterns before execution.

This creates several security risks:
- **SQL Injection**: Malicious queries could exploit string concatenation or bypass parameterization
- **Command Execution**: Attackers could use database-specific functions like `xp_cmdshell` (SQL Server), `COPY ... FROM PROGRAM` (PostgreSQL), or `sys_exec` (MySQL) to execute system commands
- **File Operations**: Queries could read/write files using `LOAD_FILE`, `INTO OUTFILE`, or similar functions
- **Arbitrary Code Execution**: MongoDB's `$where` operator and Redis's `EVAL` command allow running arbitrary JavaScript/Lua code
- **Denial of Service**: Excessively long or deeply nested queries could overwhelm the database

Even with parameterized queries, defense-in-depth is critical for a feature that directly exposes database access through a web interface.

## What Changed

Added a comprehensive query validation layer that acts as a security gate before any query reaches the database:

**New file: `backend/db-client/query-validator.ts`**
- `validateQuery()` - Main entry point that routes to driver-specific validators
- `validateSqlQuery()` - Validates SQL queries (MySQL, PostgreSQL, SQLite) against 15+ dangerous patterns and 10+ suspicious patterns
- `validateMongoQuery()` - Blocks MongoDB operators that allow arbitrary JavaScript execution (`$where`, `$function`, `mapReduce`)
- `validateRedisCommand()` - Blocks dangerous Redis commands (`EVAL`, `FLUSHALL`, `CONFIG`, `SHUTDOWN`, etc.)

**Updated: `backend/ws/db-client/query.ts`**
- Integrated validation into both `db-client:execute-read` and `db-client:execute-write` handlers
- Queries are validated before execution; unsafe queries are rejected with clear error messages
- Warnings (suspicious but not dangerous patterns) are logged for monitoring
- Validation failures are recorded in query history for audit purposes

**New file: `backend/db-client/query-validator.test.ts`**
- 31 comprehensive test cases covering all validators
- Tests for SQL injection patterns, command execution attempts, file operations, and more
- Tests for MongoDB and Redis-specific attack vectors
- Tests for edge cases like excessively long queries and deep nesting

## Validation

✅ **All tests pass** (31/31)
```bash
bun test backend/db-client/query-validator.test.ts
```

✅ **Type checking passes**
```bash
bun run check
```

✅ **Linting passes**
```bash
bun run lint
```

The validation layer is designed as defense-in-depth and works alongside existing security measures. It catches obvious attack patterns while still allowing legitimate queries to execute normally.

## Risk Assessment

**Low risk** - This is a pure addition that doesn't modify existing query execution logic. The validator only blocks queries that match known dangerous patterns. Legitimate queries continue to work as before.

If a legitimate query is accidentally blocked, users will see a clear error message explaining which pattern triggered the validation failure, making it easy to diagnose and report false positives.

### DIFF
--- a/backend/db-client/query-validator.test.ts
+++ b/backend/db-client/query-validator.test.ts
@@ -1,0 +1,191 @@
+import { describe, expect, test } from 'bun:test';
+import { validateQuery, validateSqlQuery, validateMongoQuery, validateRedisCommand } from './query-validator';
+
+describe('SQL Query Validation', () => {
+	test('should allow safe SELECT queries', () => {
+		const result = validateSqlQuery('postgres', 'SELECT * FROM users WHERE id = $1', [123]);
+		expect(result.safe).toBe(true);
+		expect(result.errors).toHaveLength(0);
+	});
+
+	test('should block xp_cmdshell attempts', () => {
+		const result = validateSqlQuery('mysql', "SELECT * FROM users; EXEC xp_cmdshell('dir')");
+		expect(result.safe).toBe(false);
+		expect(result.errors.length).toBeGreaterThan(0);
+	});
+
+	test('should block COPY FROM PROGRAM attempts', () => {
+		const result = validateSqlQuery('postgres', "COPY users FROM PROGRAM 'cat /etc/passwd'");
+		expect(result.safe).toBe(false);
+		expect(result.errors.length).toBeGreaterThan(0);
+	});
+
+	test('should block INTO OUTFILE attempts', () => {
+		const result = validateSqlQuery('mysql', "SELECT * FROM users INTO OUTFILE '/tmp/users.txt'");
+		expect(result.safe).toBe(false);
+		expect(result.errors.length).toBeGreaterThan(0);
+	});
+
+	test('should block stacked DROP queries', () => {
+		const result = validateSqlQuery('postgres', 'SELECT * FROM users; DROP TABLE users');
+		expect(result.safe).toBe(false);
+		expect(result.errors.length).toBeGreaterThan(0);
+	});
+
+	test('should block UNION-based information_schema injection', () => {
+		const result = validateSqlQuery('mysql', 'SELECT * FROM users UNION SELECT * FROM information_schema.tables');
+		expect(result.safe).toBe(false);
+		expect(result.errors.length).toBeGreaterThan(0);
+	});
+
+	test('should warn about inline string values', () => {
+		const result = validateSqlQuery('postgres', "SELECT * FROM users WHERE name = 'admin'");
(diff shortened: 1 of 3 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks for the careful writeup here — the per-driver breakdown of attack classes (command execution, file ops, `$where`/`EVAL` code exec) is genuinely thorough, and the 31-case test suite shows real diligence. I'm going to close this one, but I want to walk through why so it doesn't read as a brush-off.

The premise is that the DB Client needs a validation layer to stop SQL injection and dangerous operations before they reach the database. The thing is, this feature's whole purpose is to let a user run arbitrary queries against a database **they own, using credentials they configured**. Connection access is owner-scoped — `requireDbClientConnectionAccess` only resolves for the connection's owner (or an admin, who already controls the instance). So:

- **There's no injection surface.** The user authors the entire query directly; the app never concatenates untrusted input into an app-built statement, and `params` already flow through the driver's parameterized execution. Nothing is being "injected into."
- **`xp_cmdshell` / `COPY … FROM PROGRAM` / Redis `EVAL` aren't an escalation.** The connection owner could run those exact statements through `psql`/`mysql`/`redis-cli` with the same credentials. The database enforces what those credentials are permitted to do. Blocking them in this one UI removes capability the owner already legitimately has, without adding a security boundary.

Separately, even setting the threat model aside, the blocklist would break legitimate use: `db-client:execute-write` exists specifically to run writes and DDL, but the patterns reject stacked DDL, `exec(`/`execute(` (stored-proc calls), and `INTO OUTFILE` exports. The greedy patterns also false-positive on string literals — `SELECT note FROM logs WHERE note = 'copy data from the program team'` trips the `COPY … FROM PROGRAM` rule. And the codebase already classifies statements with comment-aware parsing in `query-executor.ts` (`classifyQuery` / `firstSqlKeyword`), with the read/write split as the actual gate — a regex layer using `\s+` is bypassable by `/**/` comments, which that existing parser was written to handle.

What would change my mind: if connections become shareable with users who are neither the owner nor an admin — a lower-privilege member running queries through someone else's stored credentials would be a real escalation surface worth hardening. Or a concrete app-side path that interpolates untrusted input into a query. Reopen here anytime if either shows up; I'd much rather walk through it together than close the door.

PR #318, opened 2026-06-04
fix(files): replace shell-based file operations with fs.promises on Windows

DeryFerd:
### Problem

A handful of file operations in this codebase were routing through `Bun.spawn` to call shell commands — `cmd /c del`, `cmd /c move`, `rm -rf`, `dir /b`, `ls -1A`, you name it. Every one of those is a small landmine. On Windows, paths with spaces or special characters can break shell parsing. On Linux, a different code path means more to maintain and more chances for something to drift apart.

This PR replaces all of those shell invocations with the corresponding `fs.promises` APIs. No more conditionals checking `process.platform` for basic file operations.

### What changed in `backend/files/file-operations.ts`

**`readdir()`** — was spawning `cmd /c dir /b /a path` on Windows or `ls -1A path` everywhere else, then scraping the output with `.trim().split(/\r?\n/)`. That approach loses metadata (file types, sizes), chokes on filenames with newlines, and is about as fragile as you'd expect. Now it calls `fs.promises.readdir` with `withFileTypes: true` and maps the names out cleanly.

**`mkdir()`** — the old implementation did something genuinely bizarre: it wrote a temporary file with `Bun.write`, then spawned `del` or `rm` to clean it up. If the process crashed between those two steps, the temp file stayed behind forever. Now it just calls `fs.promises.mkdir` with `{ recursive: true }`. One call, no leftover artifacts, no shell.

**`rename()`** — was spawning `cmd /c move` on Windows and `mv` on Unix. `fs.promises.rename` handles both natively and handles paths with spaces without any quoting gymnastics.

**`unlink()`** — spawned `cmd /c del /f /q` or `rm -f`. Replaced with `fs.promises.unlink`.

**`rm()`** — spawned `cmd /c rmdir /s /q` or `rm -rf`. Replaced with `fs.promises.rm` using `{ recursive: true, force: true }`.

### What changed in `backend/files/path-browsing.ts`

Same `readdir()` replacement — shell-based directory listing replaced with `fs.promises.readdir`.

### Why this matters beyond code cleanup

Three reasons:

1. **Shell escaping bugs are real.** If someone uploads a file or names a project with a name like `foo & bar`, the `cmd /c` variant will interpret `&` as a command separator. That's not theoretical — path-browsing exposes user-controlled paths through the file picker.

2. **Platform branches lie dormant.** The `if (process.platform === 'win32')` branches look like they handle differences, but nobody tests both paths equally. One branch eventually rots. Removing the branching removes the risk.

3. **Error messages from shell commands are terrible.** A failed `cmd /c del` gives you nothing useful. `fs.promises` throws proper `ENOENT`, `EPERM`, `EBUSY` errors that are debuggable and traceable.

### What I did not change

There's still one place in the same file (`rmdir`) that uses the old shell-based approach with escaped paths — it's called from a different flow and touches a different set of paths, so I left it out of scope. Can be cleaned up separately.

### Validation

- `bun run check` — passes (the 2 `sharp` errors are pre-existing on `main` too)
- `bun run lint` — clean
- `bun build --no-bundle backend/files/file-operations.ts` — compiles fine
- `bun build --no-bundle backend/files/path-browsing.ts` — compiles fine

### DIFF
--- a/backend/files/file-operations.test.ts
+++ b/backend/files/file-operations.test.ts
@@ -38,6 +38,23 @@ describe('file-operations — shell-special filenames', () => {
 		await expect(stat(filePath)).rejects.toThrow();
 	});
 
+	test('deleteOperation removes an empty directory', async () => {
+		const dirPath = join(testDir, 'empty-dir');
+		await mkdir(dirPath, { recursive: true });
+
+		await expect(deleteOperation(dirPath)).resolves.toMatchObject({ message: 'Directory deleted successfully' });
+		await expect(stat(dirPath)).rejects.toThrow();
+	});
+
+	test('deleteOperation force-removes a non-empty directory', async () => {
+		const dirPath = join(testDir, 'non-empty-dir');
+		await mkdir(dirPath, { recursive: true });
+		await Bun.write(join(dirPath, 'child.txt'), 'contents');
+
+		await expect(deleteOperation(dirPath, true)).resolves.toMatchObject({ message: 'Directory deleted successfully' });
+		await expect(stat(dirPath)).rejects.toThrow();
+	});
+
 	test('renameOperation renames a file whose name contains shell metacharacters', async () => {
 		const src = join(testDir, SPECIAL_NAME);
 		const dest = join(testDir, 'safe-name.txt');
--- a/backend/files/file-operations.test.ts
+++ b/backend/files/file-operations.test.ts
@@ -1,0 +1,58 @@
+import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test';
+import { rm, mkdir, stat } from 'node:fs/promises';
+import { join } from 'node:path';
+import { tmpdir } from 'node:os';
+
+// Stub settingsQueries so the module under test never touches a real database.
+await mock.module('../database/queries', () => ({
+	settingsQueries: {
+		get: () => null,
+		getAll: () => [],
+		set: () => {},
+		delete: () => {}
+	}
+}));
+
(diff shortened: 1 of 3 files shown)


ArgaFairuz:
Thanks @DeryFerd — catching that `foo & bar` breaks shell argument parsing on both platforms was the right instinct, and the threat model in the description made the audit easy. Built on top with three small additions pushed to this branch.

Moved `fsMkdir` into the static top-level import at `file-operations.ts:2` (it was acquired via `await import()` inside the function body, inconsistent with the rest of the PR's own import style). Removed the now-dead `const { spawn } = Bun` destructuring on the same file. Aligned `path-browsing.ts` to `node:fs/promises` to match. Also added a regression test at `backend/files/file-operations.test.ts` pinning the shell-metacharacter boundary — `deleteOperation`, `renameOperation`, and `createDirectoryOperation` each called with a `foo & bar`-style name; all three green locally. `bun run check` / `bun run lint` pass clean. Synced with `main`.

ArgaFairuz:
Follow-up fix: `rmdir` was delegating to `rm(path, undefined)` which called `fsRm(path, { recursive: false, force: true })` — Bun 1.3.14 on macOS triggers `EFAULT` when `fs.promises.rm` is called on a directory without `recursive: true` (different syscall path from files). Fixed by routing empty-directory deletion through `fsRmdir` directly at `file-operations.ts:65`; force-delete via `fsRm({ recursive: true })` is unaffected. Added two more test cases (empty-dir delete and force non-empty delete) — all 5 green. `bun run check` / `bun run lint` clean.

PR #319, opened 2026-06-04
fix(files): cleanup orphaned partial uploads from failed transfers

DeryFerd:
### Problem

When you upload a file through the web UI, the server writes it to a temporary path like `<filename>.<uuid>.partial` first. Once the upload finishes, it renames that partial file to its final name. This is a pretty standard pattern — it prevents readers from seeing a half-written file.

The problem: if something goes wrong mid-upload — the browser tab closes, the network drops, the server restarts — that `.partial` file stays on disk. Forever. There was no mechanism to find and clean those up. Over time, especially on a busy instance with large file transfers, those orphans add up.

This PR introduces a lightweight tracker that knows about every active partial file and periodically sweeps any that have been sitting around for too long.

### The new file: `backend/http/upload-temp-cleanup.ts`

A class called `UploadTempCleanup` keeps a `Map<string, number>` of active temp paths and their creation timestamps. It exposes four methods:

- **`register(tempPath)`** — called when a new partial file is created. Records the path and current timestamp.
- **`deregister(tempPath)`** — called when the upload completes or when the cleanup handler runs. Removes the path from the map.
- **`start()`** — kicks off a `setInterval` that runs cleanup every 60 minutes. Tied into the server startup sequence so it only starts after the database is ready.
- **`dispose()`** — clears the interval for graceful shutdown.

The cleanup pass iterates over registered paths. Any file older than 2 hours gets deleted (`fs.promises.unlink`) and removed from the map. If deletion fails (file already gone, permissions issue), it still removes the entry — no point keeping a dead reference.

The constants:

```
CLEANUP_INTERVAL_MS = 60 * 60_000   // check every hour
TEMP_FILE_AGE_MS    = 2 * 60 * 60_000  // 2-hour grace period
```

Two hours gives plenty of room for large or slow uploads to finish, even on spotty connections.

### Changes to `backend/http/files-upload.ts`

Three lines added:

1. After constructing the `.partial` path: `uploadTempCleanup.register(tempPath)` so the tracker knows about it.
2. In the `cleanupTemp` error handler (called when upload fails or client disconnects): `uploadTempCleanup.deregister(tempPath)` — removes it from the tracker since we are about to delete the file anyway.
3. On successful upload, right before `rename`: `uploadTempCleanup.deregister(tempPath)` — file is no longer partial, no longer needs tracking.

### Changes to `backend/index.ts`

One line: `uploadTempCleanup.start()` after the database initializes, alongside the existing `sessionCleanupScheduler.start()`. The cleanup only runs when the server is fully operational.

### Design notes

I chose an in-memory registry (not a database table or filesystem scan) for a few reasons:

- **Temp files are ephemeral by nature.** Writing their locations to a persistent table felt wrong — you'd be accumulating rows for files that might not exist after a restart.
- **A filesystem scan for `*.partial` patterns every hour is wasteful** on large projects with many files, and risks matching files that happen to end in `.partial` but aren't upload temps.
- **The registry is lossy across restarts**, but that's fine. Orphaned partials from a previous server session become just another file on disk. They don't cause any harm other than taking up space. If someone really needs to reclaim that space, they can manually delete `*.partial` files — or we can add a startup scan later if it becomes a real issue.

### Validation

- `bun run check` — passes (same 2 pre-existing `sharp` errors)
- `bun run lint` — clean
- `bun build --no-bundle backend/http/upload-temp-cleanup.ts` — compiles fine


ArgaFairuz:
Hi @DeryFerd, the registry-based approach maps cleanly onto `sessionCleanupScheduler`, and the reasoning for keeping this in-memory over a DB table is well-argued. I built on top with a few fixes — everything has been pushed to this branch.

## Summary

Resolved a structural conflict with #279, added missing `deregister` calls on direct-unlink paths that bypass `cleanupTemp()`, and wired `dispose()` into graceful shutdown.

## Why

PR #279 (merged after this PR branched) replaced `rename(tempPath, resolvedFinal)` with `link(tempPath, resolvedFinal)` followed by `await unlink(tempPath)` at `backend/http/files-upload.ts:224`. The `deregister` placed before `rename` had no anchor in the merged code.

Beyond the conflict, four error paths that delete `tempPath` inline bypass `cleanupTemp()` and were missing `deregister` calls, leaving stale registry entries until the next cleanup pass fires and hits ENOENT:

- `backend/http/files-upload.ts:181` — `written !== fileSizeParam` guard
- `backend/http/files-upload.ts:188` — second `validateFileSize` check
- `backend/http/files-upload.ts:208` — `link()` EEXIST branch
- `backend/http/files-upload.ts:253` — outer `catch` block

## Changes

**`backend/http/files-upload.ts`:**
- Repositioned `deregister` to just before `await unlink(tempPath)` on the success path — this is where the hard-link is complete and the temp file is about to be removed from disk.
- Added `uploadTempCleanup.deregister(tempPath)` to the four direct-unlink paths above.

**`backend/index.ts`:**
- Added `uploadTempCleanup.dispose()` to `gracefulShutdown` alongside `sessionCleanupScheduler.dispose()` (`backend/index.ts:218`) — same wiring the other interval schedulers follow.

## Notes
- Kept scope tight; cleanup logic and constants unchanged.
- `bun run check` and `bun run lint` pass clean. Synced with `main`.

PR #324, opened 2026-06-04
fix(auth): encrypt engine account credentials at rest with AES-256-GCM

DeryFerd:
## Problem

Engine account credentials — API keys, OAuth tokens, session cookies — were stored as plaintext in the `engine_accounts.credential` column. Anyone with read access to the SQLite database file (a backup, a disk snapshot, a compromised container) could extract every credential for every engine the server had ever configured.

This PR adds transparent AES-256-GCM encryption at rest for all credential data.

### Key management

The encryption key lives at `~/.clopen/credential-key` as a hex-encoded 32-byte value. On first access, the file is generated automatically using `crypto.getRandomValues`. The user can also set `CLOPEN_CREDENTIAL_KEY` as an environment variable to override the file — useful for containerized deployments where the key should come from a secret manager.

The key is loaded once and cached in memory (`cachedKey`), so repeated encrypt/decrypt calls do not hit the filesystem again.

### Encryption scheme (`backend/auth/credential-crypto.ts`)

- **Algorithm:** AES-GCM (256-bit)
- **IV:** 12 random bytes per encryption, stored alongside the ciphertext
- **Auth tag:** 16-byte GCM authentication tag, appended to the ciphertext
- **Output format:** `v1:{base64(iv)}:{base64(authTag)}:{base64(ciphertext)}`

The `v1:` prefix lets the system distinguish encrypted credentials from legacy plaintext ones (which might be anything from a raw API key like `sk-...` to a JSON blob). The `isEncrypted()` helper checks for this prefix before deciding whether to decrypt.

The `encryptCredential()` and `decryptCredential()` functions are the public API — simple strings in, strings out.

### Integration into the query layer (`backend/database/queries/engine-queries.ts`)

Six methods in `engineQueries` were changed from synchronous to asynchronous:

- `getAccount` — decrypts the credential before returning
- `getAccountsByProvider` — decrypts all accounts in the result set
- `getActiveAccount` — decrypts on read
- `getActiveAccountForEngine` — decrypts on read
- `createAccount` — encrypts the credential before inserting; returns the plaintext credential in the response object (so the caller never sees the ciphertext)
- `updateAccountCredential` — encrypts before writing
- `getProvidersWithAccounts` — now awaits the per-provider account queries

Two helper functions — `decryptAccount` and `decryptAccounts` — keep the boilerplate out of each query method. They skip decryption if the credential is not encrypted (backward compatibility).

### Caller updates (20 files)

Every call to these methods that was already in an `async` function needed `await` added. The changes are mechanical but widespread because the query methods were used throughout the WebSocket handlers, stream adapters, and environment setup code:

- All engine accounts handlers (`claude/accounts.ts`, `codex/accounts.ts`, `copilot/accounts.ts`, `qwen/accounts.ts`, `opencode/providers.ts`)
- All engine status handlers (`claude/status.ts`, `codex/status.ts`, `copilot/status.ts`, `qwen/status.ts`)
- All stream adapters (`claude/stream.ts`, `codex/stream.ts`, `copilot/stream.ts`, `qwen/stream.ts`)
- All environment setup (`claude/environment.ts`, `qwen/environment.ts`)
- OpenCode config and server (`opencode/config.ts`, `opencode/server.ts`)
- Codex credential snapshot (`codex/credential.ts`)

### Migration (`backend/database/migrations/040_encrypt_engine_credentials.ts`)

Migration 040 reads every row from `engine_accounts`, checks if it is already encrypted (skips if so), encrypts the plaintext credential, and writes it back. The `down` migration is intentionally a no-op — there is no safe way to reverse encryption without keeping a plaintext backup.

The migration runner was also updated to support async migrations. The `Migration` interface now allows `void | Promise<void>` for `up` and `down`, and both `runMigrations()` and `rollbackMigration()` `await` the calls.

## Why AES-GCM

AES-GCM provides authenticated encryption — if someone tampers with the ciphertext, decryption fails with an error rather than returning garbage. This is important because credentials decrypted to garbage would cause confusing auth failures that are hard to debug. GCM also handles the IV correctly without needing a separate padding scheme.

## What this does not cover

- **Key rotation** — there is no mechanism to rotate the key and re-encrypt all credentials. The migration encrypts existing credentials once on first run. If the key is lost, credentials cannot be recovered.
- **In-memory exposure** — credentials are decrypted on read and held in memory during the request. They are not cached beyond that.
- **Network encryption** — this is about storage, not transport. TLS is a separate concern.

## Risk

The biggest risk is a failed migration leaving credentials in an inconsistent state. The migration handles this by:
1. Checking `isEncrypted()` before attempting encryption, so it is idempotent
2. Running within a single-threaded context (SQLite is single-writer), so there is no race condition

If the migration or encryption fails mid-way, affected credentials become unreadable (GCM auth tag mismatch). In practice, this would mean users need to re-enter their API keys for the affected accounts. A backup of the database before migration is strongly recommended.

## Validation

- `bun run check` — passes (2 pre-existing `sharp` errors, same as `main`)
- `bun run lint` — clean
- `bun build --no-bundle` compiles all affected files individually

### DIFF
--- a/backend/auth/credential-crypto.ts
+++ b/backend/auth/credential-crypto.ts
@@ -1,0 +1,114 @@
+import { mkdir, readFile, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { homedir } from 'node:os';
+import { debug } from '$shared/utils/logger';
+
+const ALGORITHM = 'AES-GCM';
+const KEY_LENGTH = 256;
+const IV_LENGTH = 12;
+const KEY_FILE = join(homedir(), '.clopen', 'credential-key');
+const CIPHER_VERSION = 'v1';
+
+let cachedKey: CryptoKey | null = null;
+
+function bytesToHex(bytes: Uint8Array): string {
+	return Array.from(bytes).map(b => b.toString(16).padStart(2, '0')).join('');
+}
+
+function hexToBytes(hex: string): Uint8Array {
+	const bytes = new Uint8Array(hex.length / 2);
+	for (let i = 0; i < hex.length; i += 2) {
+		bytes[i / 2] = parseInt(hex.slice(i, i + 2), 16);
+	}
+	return bytes;
+}
+
+function base64Encode(bytes: Uint8Array): string {
+	return btoa(String.fromCharCode(...bytes));
+}
+
+function base64Decode(str: string): Uint8Array {
+	const binary = atob(str);
+	const bytes = new Uint8Array(binary.length);
+	for (let i = 0; i < binary.length; i++) {
+		bytes[i] = binary.charCodeAt(i);
+	}
+	return bytes;
+}
+
+async function loadOrCreateKey(): Promise<CryptoKey> {
+	const envKey = process.env.CLOPEN_CREDENTIAL_KEY;
+	if (envKey) {
+		const raw = hexToBytes(envKey);
+		const rawArr = new ArrayBuffer(raw.byteLength);
+		new Uint8Array(rawArr).set(raw);
+		return await crypto.subtle.importKey('raw', rawArr, ALGORITHM, false, ['encrypt', 'decrypt']);
+	}
+
+	let raw: Uint8Array;
+	try {
+		const stored = await readFile(KEY_FILE, 'utf-8');
+		raw = hexToBytes(stored.trim());
+	} catch {
+		raw = crypto.getRandomValues(new Uint8Array(32));
+		await mkdir(join(homedir(), '.clopen'), { recursive: true });
+		await writeFile(KEY_FILE, bytesToHex(raw), 'utf-8');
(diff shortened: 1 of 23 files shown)


ArgaFairuz:
Appreciate the thoroughness here — the AES-GCM implementation itself is clean and
the `v1:` prefix for legacy detection is a nice touch.

The threat model doesn't hold up, though. The key lives at
`~/.clopen/credential-key`, in the same directory as the database. A disk
snapshot, a container compromise, or a directory backup — the three cases the
description names — all capture the key alongside the ciphertext.

There's also a practical blocker: migration slot 040 is already
`040_create_mcp_servers_table.ts`, and turning six query methods async ripples
through 20 files in the engine layer. That's a lot of surface for the protection
it currently buys.

Closing. If we revisit this, the key needs to come from somewhere the database
backup doesn't reach.

PR #343, opened 2026-06-13
feat(history): add copy and compress session actions menu

rust142:
# Add session actions menu with Copy and Compress options in History modal

## Summary

This PR introduces a session actions menu in the History modal, allowing users to copy a session's chat to the clipboard or compress a session with AI into an ultra-terse summary.

## Changes

- Added an ellipsis actions menu button for each session in the history list.
- Implemented "Copy" action to copy the full chat transcript to the clipboard in a readable format.
- Implemented "Compress" action to generate a summarized session using an AI "caveman" prompt and start a new session.

## Preview

image

### DIFF
--- a/frontend/components/history/HistoryModal.svelte
+++ b/frontend/components/history/HistoryModal.svelte
@@ -2,9 +2,11 @@
 	import { untrack } from 'svelte';
 	import { sessionState, setCurrentSession, removeSession, reloadSessionsForProject } from '$frontend/stores/core/sessions.svelte';
 	import { projectState } from '$frontend/stores/core/projects.svelte';
+	import { setCurrentView, appState } from '$frontend/stores/core/app.svelte';
 	import { addNotification } from '$frontend/stores/ui/notification.svelte';
 	import ws from '$frontend/utils/ws';
 	import type { ChatSession } from '$shared/types/database/schema';
+	import type { UnifiedMessage } from '$shared/types/unified';
 	import Icon from '$frontend/components/common/display/Icon.svelte';
 	import AvatarBubble from '$frontend/components/common/display/AvatarBubble.svelte';
 	import Modal from '$frontend/components/common/overlay/Modal.svelte';
@@ -14,6 +16,8 @@
 	import { userStore } from '$frontend/stores/features/user.svelte';
 	import { debug } from '$shared/utils/logger';
 	import { modelStore } from '$frontend/stores/features/models.svelte';
+	import { chatService } from '$frontend/services/chat/chat.service';
+	import { setChatScroll } from '$frontend/stores/features/chat-workspace.svelte';
 
 	interface Props {
 		isOpen: boolean;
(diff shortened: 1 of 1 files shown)


ArgaFairuz:
Thanks @rust142 — the content-block handling in `formatMessageBody` is genuinely careful: it discriminates on every block type, falls back cleanly to `[No content]`, and the loading/`disabled` states plus the success/error notifications make the menu feel finished. The Copy action in particular is solid on its own.

I'd like to talk through the Compress design before this lands, though, because I don't think the current mechanism delivers what it promises.

## Compress doesn't actually compress

`compressSessionWithAI` sends the entire HEAD transcript as the *user message* of the new session and then asks the model to summarize it. So the new session's context ends up holding `[full transcript] + [terse handoff]` — larger than the source, not smaller. The saving only ever materializes if later auto-compaction drops the original prompt, or if the user manually copies the handoff into a third, clean session. As built it's closer to "copy HEAD into a new session and ask for a summary" than a real compress.

A compress that holds up on tokens needs to seed the new session with *only* the handoff, not its input — e.g. run the summarization in a throwaway context, capture the assistant output, then create the new session whose first message is just that handoff. A server-side variant would work too.

## The transcript is both lossy and noisy

This compounds the above, because the material we feed the summarizer is the wrong shape:

- **Tool inputs are dropped.** Tool calls render as `[Tool: <name>]` only, discarding `block.input` — which is exactly the file paths, commands, and URLs the caveman prompt asks for. `[Tool: navigate]` without the URL is close to useless for a handoff.
- **Call and result are split.** The enriched result already lives on the call block (`ToolUseBlock.result`, `shared/types/unified/tool.ts:435`); relying instead on the separate `tool_result` user message means each call and its result land in different sections for the model to re-correlate.
- **Noise goes in verbatim.** Empty `[No content]` assistant turns, `Screenshot captured successfully` confirmations, and full DOM dumps from `analyze_dom` are all forwarded raw with no truncation — so the prompt is huge *and* missing the useful bits.

## Diverges from the shared-session flow

`compressSessionWithAI` is the only frontend caller of `sessions:create` — every other new-session path goes through `sessions:get-shared` (`frontend/stores/core/sessions.svelte.ts:142`). That handler broadcasts `sessions:session-available` to the project (`backend/ws/sessions/crud.ts:185`); `sessions:create` does not, so the caveman session won't appear for collaborators and it sidesteps the one-active-shared-session model. Worth either reusing the shared flow or justifying the divergence.

## Minor

- `chatService.sendMessage` early-returns on `appState.isLoading` (`frontend/services/chat/chat.service.ts:355`). If switching to the freshly created session ever flips `isLoading`, the caveman prompt is silently dropped and the user lands on an empty session with no error — worth a guard.
- HEAD-only `messages:list` (no `include_all`) is the right call here; flagging only so it stays a conscious choice.

Happy to pair on reshaping Compress if that helps. The Copy half could also land on its own if you'd rather split the two.

Could you take a look by June 21, 2026? If you can't respond by then, I'll close this PR as auto-stale — you can reopen anytime once you're back.

PR #345, opened 2026-06-13
fix(files): improve cross-platform mount detection and FolderBrowser filtering

rust142:
# Refine mount detection and system location filtering in FolderBrowser

## Summary

This PR improves the detection and filtering of mount points and system locations in both the backend and the FolderBrowser component, resulting in more accurate and user-friendly folder navigation mac platforms.

## Changes

- Added better directory existence/stat helpers in backend for mount detection.
- Exclude likely macOS app volumes (e.g. DMGs) from the list of selectable volumes.
- Improved logic to filter and display only relevant/usable system locations in FolderBrowser.
- Enhanced highlighting and selection logic for home, drives, and quick-access locations.
- Avoids auto-selection of the synthetic "drives" root as a valid folder.
- Updated UI behavior for dynamic location buttons and improved styling for active locations.

### DIFF
--- a/backend/files/path-browsing.ts
+++ b/backend/files/path-browsing.ts
@@ -18,15 +18,14 @@ async function statIfExists(path: string): Promise<Awaited<ReturnType<ReturnType
 	}
 }
 
+// Returns true when every visible entry is a .app bundle or the "Applications" alias —
+// the signature of a DMG/installer mount, not a general-purpose volume.
 async function isLikelyMacAppVolume(volumePath: string): Promise<boolean> {
 	try {
 		const entries = await readdir(volumePath);
-		const hasApplicationsAlias = entries.includes('Applications');
-		const appBundleCount = entries.filter(entry => entry.endsWith('.app')).length;
-
-		// Typical DMG/app mounts expose one or two app bundles plus an
-		// Applications shortcut. Hide those from the project picker.
-		return hasApplicationsAlias && appBundleCount > 0 && entries.length <= 8;
+		const visible = entries.filter(e => !e.startsWith('.'));
+		if (visible.length === 0) return false;
+		return visible.every(e => e.endsWith('.app') || e === 'Applications');
 	} catch {
 		return false;
 	}
@@ -45,6 +44,7 @@ export interface PathBrowseData {
 		type: 'file' | 'directory';
 		path: string;
 		modified: string;
+		kind?: 'home' | 'root' | 'volume' | 'drive';
 		size?: number;
 		extension?: string;
 	}>;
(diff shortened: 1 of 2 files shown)


ArgaFairuz:
Hi @rust142, thanks a lot for this contribution! The `statIfExists` refactor was the right call, and catching DMG false-positives in the drives view is a real UX win. I built on top of your commit with a few additional fixes — everything has been pushed to this branch.

## Summary

Extended mount detection to be fully cross-platform (Linux `/run/media` support, platform-specific branch per OS), replaced `isLikelyMacAppVolume`'s magic constant with a content-based check, introduced a `kind` field so the frontend filters by semantics rather than display-name strings, and fixed two UX issues in the quick-access bar.

## Why

Five issues surfaced in the audit:

- `isLikelyMacAppVolume` at `backend/files/path-browsing.ts` used `entries.length <= 8` as the DMG cutoff — a magic constant that hides small USB drives and misses large DMG archives. A content-based check is tighter and self-documenting.
- `isUsefulSystemLocation` at `frontend/components/common/form/FolderBrowser.svelte` filtered by `location.name === 'Home Directory'`, coupling the frontend to a string literal in the backend. If that name ever changes, the filter silently breaks with no type error.
- `isQuickAccessActive('home')` used `currentPath.startsWith('/Users/')` as a proxy for the home directory, which lights up the Home button when navigating any `/Users/…` path — not just the actual home.
- The Linux path lacked `/run/media/{user}/{volume}` support, where systemd-based distros (Ubuntu, Fedora, Arch) mount removable media.
- When in the drives view, "Home Directory" appeared in the content list despite the dedicated Home button already covering it — and the System button sat between Home and the volume shortcuts instead of at the end where it reads as the "all locations" overview.

## Changes

**`backend/files/path-browsing.ts`:**
- Replaced `isLikelyMacAppVolume`'s `entries.length <= 8` threshold with `visible.every(e => e.endsWith('.app') || e === 'Applications')` — fires on content, not count.
- Rewrote `handleUnixMountPoints` with platform-specific branches: macOS gets home + `/Volumes/*` (DMGs skipped); Linux gets home + root + `/mnt/*` + `/media/*` + `/run/media/{user}/*`; other Unix gets home + root + `/mnt/*` + `/media/*`. Removed `/usr`, `/var`, `/opt`, `/tmp` — never useful in a project picker.
- Added `kind?: 'home' | 'root' | 'volume' | 'drive'` to `PathBrowseData` children and tagged every item in `handleUnixMountPoints` and `handleWindowsDrives` accordingly.

**`frontend/components/common/form/FolderBrowser.svelte`:**
- Added `kind?` to `FileItem` interface.
- Replaced `isUsefulSystemLocation`'s name-string filter with `kind`-based: `quickAccessLocations = availableDrives.filter(d => d.kind !== 'home')`.
- Added `homePath` derived from `availableDrives.find(d => d.kind === 'home')?.path` — actual home path from the backend, no client-side guessing.
- Fixed `isQuickAccessActive('home')` to use `pathsEqual(currentPath, homePath)` — exact match, not a prefix heuristic.
- Fixed `filteredItems` to hide `kind === 'home'` items when `currentPath === 'drives'` — Home Directory has a dedicated button, no reason to duplicate it in the list.
- Reordered quick-access buttons to **Home → [volumes/drives] → System**, so System reads as the "all locations" overview rather than a peer of specific destinations.
- Fixed dynamic button icon logic to branch on `location.kind` instead of path prefix checks — works correctly for `/run/media` volumes and any future mount shapes.
- Disabled the "up" button when `currentPath === 'drives'` (was only disabled at `/`).
- `manualPath` effect: shows `homePath || '/'` instead of hardcoded `'/'` when on the drives view.

## Notes
- `bun run check` and `bun run lint` pass clean. Synced with `main`.

ArgaFairuz:
Found an additional regression in Windows testing after the previous push: navigating into any drive (e.g. clicking the `C:` button) produced an empty listing — "No folders found / This directory doesn't contain any subdirectories."

**Root cause.** Windows treats a bare drive letter (`C:`) as a *drive-relative path* that resolves to the drive's current working directory, not its root. Path normalisation and manual entry can both strip the trailing separator, so when `handlePathBrowsing` received `C:` and called `join('C:', item)` for each child, it produced `C:Users` instead of `C:\Users`. Every `stat()` call on those drive-relative child paths failed silently, leaving the children array empty.

**Fix** — `backend/files/path-browsing.ts` — pushed in `fde4f83`. After the `targetPath` is resolved but before `stat()` is called, a guard appends `\\` whenever the path matches `^[A-Za-z]:$`:

```ts
if (/^[A-Za-z]:$/.test(targetPath)) {
    targetPath += '\\';
}
```

This covers both manual entry (`C:` typed without the separator) and any normalisation path that strips the trailing backslash.

PR #346, opened 2026-06-13
feat(git): add checkout commit and open commit on remote actions in history

rust142:
## Summary

This PR introduces two new actions to the Git log panel: checking out a commit (detached HEAD) and opening a commit on the remote repository (e.g., GitHub, Bitbucket).

 ## Changes

 - Added API and backend logic for checking out a specific commit by hash.
 - Exposed a new "Checkout commit" action in the Git log UI, with confirmation dialog.
 - Added logic to detect the remote URL and build links to a commit page (supports major hosts).
 - Added "Open commit on remote" action in the Git log UI; opens the commit in default browser.
 - Improved the log row's UI to show new action buttons on hover/focus.
 
 ## Preview
 
image

### DIFF
--- a/backend/git/git-service.ts
+++ b/backend/git/git-service.ts
@@ -317,6 +317,14 @@ export class GitService {
 		}
 	}
 
+	async checkoutCommit(cwd: string, commitHash: string): Promise<void> {
+		assertSafeGitRevish(commitHash, 'commit hash');
+		const result = await execGit(['checkout', commitHash], cwd);
+		if (result.exitCode !== 0) {
+			throw new Error(`git checkout failed: ${result.stderr}`);
+		}
+	}
+
 	async deleteBranch(cwd: string, name: string, force = false): Promise<void> {
 		assertSafeGitRevish(name, 'branch name');
 		const flag = force ? '-D' : '-d';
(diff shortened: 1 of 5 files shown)


ArgaFairuz:
Thanks @rust142 — nice work on this one. The checkout flow matches the existing `switchBranch` pattern exactly, and the remote URL builder handles the common hosts cleanly. Checked adjacent call sites in `backend/ws/git/` and `frontend/components/git/` — all consistent. `bun run check` / `bun run lint` green locally.

I resolved the merge conflict in `GitPanel.svelte` (main updated `createBranch` return type to `Promise<boolean>`) and fixed a bug where the commit log wasn't reloaded after checkout — both `checkoutCommit` and `switchBranch` were missing `loadLog(true)` after `loadAll()`, so switching to a different commit/branch left the old history visible.

Merging.

ArgaFairuz:
Heads-up for the next person: I pushed a follow-up to disable the "Open commit on remote" button for unpushed commits (backend tags `isPushed` via `git rev-list <ref> --not --remotes`, UI disables on `isPushed === false`), but I'm pausing it.

The problem: `--not --remotes` only sees remote-tracking refs in the local clone. When you push a commit to a contributor's fork branch, that fork usually isn't a tracked remote here, so it's flagged "not pushed" even though it's live on GitHub (and the origin URL would resolve fine via the fork network). Offline there's no clean way to tell "pushed to an untracked fork" from "purely local" — so I'd rather settle the intended behavior first. Leaving it out of this PR.

PR #347, opened 2026-06-13
feat(files): add cross-platform reveal-in-file-manager in file context menu

rust142:
# Add "Reveal in Finder" option for macOS in file context menu

## Summary

This PR introduces a "Reveal in Finder" option to the file context menu for macOS users, allowing them to quickly locate files in Finder directly from the UI.

## Changes

- Added backend support for a `files:reveal-in-finder` HTTP endpoint, which executes the Finder reveal only on macOS.
- Updated the file context menu to include a "Reveal in Finder" button, visible only when running on macOS.
- Passed `isMacOS` as a prop throughout the file tree components to control menu visibility.

## Preview

image


ArgaFairuz:
Hi @rust142, thanks for this PR — the macOS-specific backend handler was well-structured, and the context menu placement followed the existing pattern. I built on top with cross-platform support and a few structural fixes.

## Summary

Extends the original macOS-only "Reveal in Finder" to work cross-platform, tightens the action wiring, and fixes the architecture.

## Changes

**Backend — `backend/ws/files/reveal.ts` (new):**
- Cross-platform reveal handler (`files:reveal-in-file-manager`):
  - `open -R` (macOS)
  - `explorer /select,` (Windows)
  - `xdg-open` on parent directory (Linux)
- Moved out of `write.ts` into its own module, same pattern as `image-edit.ts`
- Registered in `backend/ws/files/index.ts`

**Frontend — `FileNode.svelte`:**
- Uses `isMac()/isWindows()/isLinux()` from existing `$frontend/utils/platform` instead of inline platform detection
- Context menu label adapts per platform ("Reveal in Finder" / "Show in Explorer" / "Show in Files")
- No `isMacOS` prop needed — eliminates prop drilling and the recursive-child gap
- Gated behind `isLocalConnection()` — only visible when accessing Clopen via `localhost` or `127.0.0.1`, since spawning a file manager on the server machine is meaningless for remote clients

**Frontend — `FileTree.svelte`:**
- Removed the `isMacOS` derivation and prop passthrough

**Frontend — `FilesPanel.svelte`:**
- Added `case 'reveal-in-file-manager'` calling `ws.http('files:reveal-in-file-manager', ...)` with error handling

**Frontend — `frontend/utils/platform.ts`:**
- Added `isLocalConnection()` for reuse across components

## Security impact

`requireFilePathAccess` is called before executing any platform command. The localhost gate prevents the button from appearing for remote users.

## Notes

- `bun run check` and `bun run lint` pass clean
- The `isLocalConnection` heuristic (`localhost` / `127.0.0.1`) is a pragmatic approximation — Clopen has no "local mode" concept. Users accessing via LAN IP or tunnel won't see the button, which is intentional since spawning processes on the server is only meaningful from the same machine.

PR #349, opened 2026-06-13
fix(terminal): preserve terminal names and avoid duplicate PTY reconnects

rust142:
# Preserve terminal tab names and prevent duplicate PTY reconnects

## Summary

This PR improves terminal session handling by preserving custom tab names across restarts and prevents duplicate PTY connections, avoiding redundant prompts and event listeners.

## Changes

- Terminal tab names are now editable and saved; they persist across sessions and reloads.
- Tab renaming UI added to `TerminalTabs.svelte`, allowing rename via double-click.
- Session persistence now saves and restores the terminal name.
- Avoids sending a duplicate initial newline to the PTY, sending it only as a fallback if the prompt doesn't display.
- Removed browser-side logic that caused duplicate PTY connections and output streams after refreshes.
- Improved PTY session cleanup to avoid multiple active listeners when switching projects or sessions.

## Preview 

image

### DIFF
--- a/backend/terminal/pty-session-manager.ts
+++ b/backend/terminal/pty-session-manager.ts
@@ -86,17 +86,7 @@ class PtySessionManager {
 		});
 
 		debug.log('terminal', `✅ PTY spawned with PID: ${pty.pid}`);
-
-		// CRITICAL: Send initial newline to trigger shell prompt display
-		// Without this, shell won't show prompt on first connection
-		setTimeout(() => {
-			try {
-				pty.write('\r');
-				debug.log('terminal', '📝 Sent initial newline to trigger prompt');
-			} catch (error) {
-				debug.error('terminal', 'Failed to send initial newline:', error);
-			}
-		}, 100);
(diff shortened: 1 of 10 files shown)


ArgaFairuz:
Thanks @rust142 — the rename work and the same-project dedup are solid, and your idempotency guards (`connectingSessions` + the `activeListeners` early-return) were the right foundation.

I pushed one fix on top: the `hasLivePty` branch in `connectToSession` reattached the WS listeners but returned before calling `terminal:create-session`, so on a project switch (A → B → A) the server never replayed the serialized scrollback — `showProjectTerminalSessions` had already reset `session.lines` to `[]`, leaving the tab blank until a refresh recreated the in-memory session and forced the full create path.

Dropping the early-return makes the replay run every time fresh listeners attach; the server handler is already idempotent (reuses the live PTY, clears stale listeners) and duplicate connects are still covered by your guards plus XTerm's `shouldConnectSession`. `bun run check` / `lint` green.

PR #403, opened 2026-07-09
fix(error-handler): redact PII from dev error responses

DeryFerd:
## Summary

`backend/middleware/error-handler.ts` returned `error.toString()` to the client in dev mode. `toString()` includes the stack trace, which carries absolute file paths and the project layout, and it also includes any string the upstream library embedded in the Error message. Real libraries do this routinely: `Bun.sql` puts the connection string in SQL errors, Node's `ENOENT` puts the full filesystem path, `ssh2` puts the auth method and host. There is no library convention that says "strip PII from the message before throwing."

The result is a real PII leak in any LAN-binding dev session (`CLOPEN_HOST=0.0.0.0`). Anyone on the same network can hit the dev server, trigger an error, and read the operator's project layout, file paths, or database credentials straight from the response body. The same body also tells them which internal services are reachable and on what ports, which is useful pre-attack reconnaissance.

Production mode already returns a hard-coded generic string, so this fix is dev-only behavior. The production path is unchanged.

## What this PR changes

Replaces `error.toString()` with `devErrorSummary(error)`, a new exported helper that returns a redacted, capped, single-line string. The full `error` object is still logged server-side via `console.error('[Error]', code, reqInfo, error)` so operators can see what actually happened. Only the client-facing body is redacted.

The summary is built in three passes:

1. Take `error.message` (not `error.toString()`, so no stack frame reaches the client).
2. Redact well-known PII patterns with regex substitution.
3. Collapse all CR/LF to a single space and cap at 200 characters.

Empty messages and nullish non-Errors collapse to a generic `"An error occurred"`. This is intentional: returning the literal string `"null"` or `"undefined"` would help an attacker probe the server's error model.

## Redaction patterns

The Error class itself does not put secrets in the message. The risk is downstream libraries that format the throw. We redact the well-known shapes and let everything else pass through (capped, newline-stripped, single line). The list is conservative, so legitimate error messages still read clearly to the operator in the browser dev tools.

| Pattern | Replacement | Catches |
|---|---|---|
| Windows absolute paths: `C:\Users\Alice\…\file.ext` | `[PATH]` | Node `ENOENT`, Bun file errors, `fs` operations |
| POSIX `/home/`, `/root/`, `/Users/`, `/var/`, `/etc/`, `/tmp/`, `/opt/` followed by any path | `[PATH]` | Same on Linux/macOS |
| `*.sqlite`, `*.sqlite3`, `*.db`, `*.sql` filenames | `[DB]` | Standalone DB files mentioned outside a longer path |
| `postgres://`, `postgresql://`, `mysql://`, `mongodb://`, `mongodb+srv://`, `redis://` followed by any URL | `[URL]` | Bun.sql connection strings, MongoDB driver, redis client, knex |
| `Bearer `, `Basic `, `Token `, `api_key=`, `api-key=`, `secret=` followed by 8+ base64/url-safe chars | `[REDACTED]` | Library-formatted upstream errors that embed the credential |

The regexes are tested in `backend/middleware/error-handler.test.ts`. The list is small enough to audit by eye; adding more patterns is easy if a real leak shows up in the field.

## Failure modes

| Scenario | Before | After |
|---|---|---|
| `ENOENT` on a user-named path, dev mode, LAN reachable | Browser receives the full `Error.toString()` including `C:\Users\Alice\…\db.sqlite` | Browser receives `[PATH]` and the rest of the message |
| Bun.sql connection error, dev mode, LAN reachable | Browser receives `postgres://admin:hunter2@db.internal:5432/app` | Browser receives `[URL]` |
| Library embeds an API key in the thrown error, dev mode, LAN reachable | Browser receives the key | Browser receives `[REDACTED]` |
| Long error message (> 200 chars) | Browser receives the full string | Browser receives the first 200 chars |
| Multi-line error (e.g. captured child-process stderr) | Browser receives raw newlines, smuggle risk for header injection | Browser receives single line, no header risk |
| Empty `Error.message` or nullish non-Error | Browser receives `""` or `"null"` | Browser receives `"An error occurred"` |
| Production mode | Hard-coded `"An error occurred"` | Unchanged |
| Any error, server log | Full `console.error(..., error)` | Unchanged, full error still logged |

The server log is the source of truth. The response body is a UX hint for the developer, not a debugging channel. Operators who need the real message go to the log, which is where the full stack and original message are preserved.

## Validation

```
bun test --isolate
…
168 pass
0 fail
681 expect() calls
```

Targeted:

- `bun test --isolate backend/middleware/error-handler.test.ts`: 10/10 pass. The tests use the same `mock.module('../utils/env', ...)` plus query-string cache-buster as the CORS tests, because `errorHandlerMiddleware` reads `SERVER_ENV.NODE_ENV` at the `default:` branch. Re-importing the module per test ensures each test sees the right env.
- Coverage:
  - plain `Error` returns its message
  - stack frame is not included in the summary
  - CR/LF are collapsed, no header can be smuggled in
  - 200-character cap holds
  - empty message and nullish non-Errors collapse to the generic string
  - Windows absolute path is replaced with `[PATH]`
  - POSIX `/home/<user>/<project>` path is replaced with `[PATH]`
  - `postgres://user:pass@host/db` connection string is replaced with `[URL]`
  - `api_key=…` embedded credential is replaced with `[REDACTED]`
  - production mode re-imports cleanly (smoke test for the module init path)

`bunx eslint backend/middleware/error-handler.ts backend/middleware/error-handler.test.ts`: clean.

## Threat model

This is a defense-in-depth fix, not a primary security control. The primary control is "do not run dev mode on a network you do not trust." If a user sets `CLOPEN_HOST=0.0.0.0` and binds the dev server to the LAN, they have already accepted that anyone on the LAN can hit the dev server. The redaction is a backstop so that a casual visitor (or a curious flatmate) cannot pivot from "I can hit the dev server" to "I now know the operator's project layout and DB host."

The redaction does not protect against:

- a determined attacker who controls a request that triggers a custom error path
- a library that throws with a PII shape the regexes do not match
- a future refactor that calls `error.stack` directly somewhere else

The full redaction is a small, auditable regex set. Adding more patterns is one PR per pattern. The defense-in-depth framing means the cost of a missed pattern is bounded: the attacker still needs LAN access to the dev server, and the operator can still see the full error in the log.

## Diff stat

```
backend/middleware/error-handler.test.ts | 153 ++++++++++++++++++++++++++++++++
backend/middleware/error-handler.ts      |  98 ++++++++++++++++-----
2 files changed, 276 insertions(+), 64 deletions(-)

### DIFF
--- a/backend/middleware/error-handler.test.ts
+++ b/backend/middleware/error-handler.test.ts
@@ -1,4 +1,5 @@
 import { beforeEach, describe, expect, mock, test } from 'bun:test';
+import { Elysia, t } from 'elysia';
 
 /**
  * Tests for the redacted error summary returned to the client.
@@ -14,7 +15,7 @@ import { beforeEach, describe, expect, mock, test } from 'bun:test';
 
 type Module = {
 	devErrorSummary: (error: unknown) => string;
-	errorHandlerMiddleware: unknown;
+	errorHandlerMiddleware: (app: Elysia) => any;
 };
 
 async function loadWithNodeEnv(env: 'development' | 'production'): Promise<Module> {
@@ -129,5 +130,47 @@ describe('devErrorSummary', () => {
 		// would fail.
 		const { errorHandlerMiddleware } = await loadWithNodeEnv('production');
 		expect(typeof errorHandlerMiddleware).toBe('function');
+	});
+});
+
+describe('VALIDATION responses', () => {
+	/**
+	 * Elysia's ValidationError.message is a JSON string that embeds the
+	 * full submitted request body (`found: <value>`) — even in production.
+	 * A field that fails validation (e.g. wrong type) can sit right next
+	 * to a field that didn't (e.g. a password), and both get echoed back.
+	 * These tests hit the middleware through a real request so they catch
+	 * a regression even if a future change reaches for `error.message`
+	 * again.
+	 */
+	async function appWithValidation(env: 'development' | 'production') {
+		const { errorHandlerMiddleware } = await loadWithNodeEnv(env);
+		return new Elysia()
+			.use(errorHandlerMiddleware)
+			.post('/login', ({ body }) => body, {
+				body: t.Object({ username: t.String(), password: t.String(), age: t.Number() })
+			});
+	}
+
+	async function postInvalid(app: Elysia) {
+		return app.handle(new Request('http://localhost/login', {
+			method: 'POST',
+			headers: { 'Content-Type': 'application/json' },
+			body: JSON.stringify({ username: 'alice', password: 'SuperSecret123!', age: 'not-a-number' })
+		}));
+	}
+
+	test('does not echo the submitted request body in development', async () => {
+		const res = await postInvalid(await appWithValidation('development'));
+		const body = await res.text();
+		expect(body).not.toContain('SuperSecret123!');
+		expect(body).not.toContain('alice');
+	});
+
+	test('does not echo the submitted request body in production', async () => {
+		const res = await postInvalid(await appWithValidation('production'));
+		const body = await res.text();
+		expect(body).not.toContain('SuperSecret123!');
+		expect(body).not.toContain('alice');
 	});
 });
--- a/backend/middleware/error-handler.test.ts
+++ b/backend/middleware/error-handler.test.ts
@@ -1,0 +1,133 @@
+import { beforeEach, describe, expect, mock, test } from 'bun:test';
+
+/**
+ * Tests for the redacted error summary returned to the client.
+ *
+ * Background: the global error handler used to return
+ * `error.toString()` in dev mode, leaking absolute file paths,
+ * stack frames, and any string the upstream library embedded in
+ * the Error. We now return only the error message, capped at 200
+ * chars and stripped of newlines, so a malicious or buggy library
+ * can't smuggle a header or dump the user's project layout into
+ * the browser.
+ */
+
+type Module = {
+	devErrorSummary: (error: unknown) => string;
+	errorHandlerMiddleware: unknown;
+};
+
+async function loadWithNodeEnv(env: 'development' | 'production'): Promise<Module> {
+	mock.module('../utils/env', () => ({
+		SERVER_ENV: {
+			NODE_ENV: env,
+			PORT: 9141,
+			PORT_FRONTEND: 9151,
+			HOST: 'localhost',
+			isDevelopment: env === 'development'
+		}
+	}));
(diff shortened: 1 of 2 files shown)


ArgaFairuz:
Hi @DeryFerd, thanks for digging into this — the redaction patterns and the 10-case test suite were thorough work.

Closing this one, but wanted to walk through why.

The fix replaces `error.toString()` with a redacted summary in dev mode (`backend/middleware/error-handler.ts:131-140`), so paths and connection strings don't reach the client.

But dev mode's whole purpose is to show everything to the person debugging. That's the tool doing its job, not a leak.

Your own threat model already says the real control is not running dev mode on a network you don't trust — not what the response body contains.

If someone sets `CLOPEN_HOST=0.0.0.0`, they've already decided that network is safe enough to share. At that point, redacting the error message doesn't change the actual exposure — anyone on that network could already reach the dev server directly.

Reopen this anytime if you find a case where the redaction still matters even with that boundary in place — happy to look again.

ArgaFairuz:
## Note on tooling

Using AI to help draft PRs is fine — the regex work and test scaffolding here are genuinely solid.

But this isn't the first time.

#402 and #403 both get the mechanics right while missing how the feature is actually used in practice.

#402 fixed the CORS origin correctly, but didn't check whether live preview still works once someone actually opens the app from a LAN IP. It doesn't, because of a browser limitation unrelated to CORS.

#403 hardened dev-mode error output correctly, but didn't check whether hiding that output actually fits how dev mode is meant to be used here.

The same pattern shows up further back, and these aren't theoretical issues. Each one only surfaced by tracing the real code path, not by reasoning on paper.

#251 and #401, months apart, both diagnosed the same setup-flow race condition. However, `bun:sqlite` calls are synchronous, so running the actual routes shows the race can't happen in either case.

#276 re-fixed a symlink-escape gap that #263 had already closed on `main`.

#277 assumed PTY sessions don't store a `userId`, when they do.

#283 added query validation against a DB Client feature whose entire purpose is running arbitrary queries against a database the user already owns.

I raised this same concern back in #227: install Bun, run the change in a real app, and verify the actual scenario before opening the PR.

Every case above would have surfaced with that same process. Open a browser, run the real race against the real routes, search `main` for existing coverage, and read the actual type definition.

Please build that verification step into your process before the next PR.

Run the change, trace it against the real code, and confirm the exact case you're fixing actually happens the way you think it does.

Happy to keep reviewing these closely. I just want that verification step to happen on your end first, since repeated misses like this cost review time on both sides.

PR #427, opened 2026-07-16
fix(git): show spinner during push progress

rust142:
Extracts spinner state outside the hover-action overlay wrapper in both nested and flat branch views, simplifying the conditional class logic. Previously the overlay container toggled between visible and hover-visible states via pointer-events and opacity classes; now the overlay consistently uses group-hover opacity while the spinner is rendered as a sibling element, making the DOM structure easier to reason about and modify.

### DIFF
--- a/frontend/components/workspace/panels/GitPanel.svelte
+++ b/frontend/components/workspace/panels/GitPanel.svelte
@@ -3658,7 +3658,7 @@ ${bodies}`;
 				</div>
 			</div>
 			{#if !branch.isCurrent}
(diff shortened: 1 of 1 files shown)


ArgaFairuz:
Hi @rust142, thanks for catching this — the push spinner disappearing mid-push because it lived inside a hover-only overlay was a real bug, and going back for the refactor pass in your second commit to tidy up the conditional structure was a good instinct.

Going to close this in favor of a broader replacement, though: the root cause isn't scoped to the push spinner. Every action button in this panel (branch switch/push/merge/delete, commit view-diff/cherry-pick/undo, remote star/edit/fetch/disconnect, stash pop/drop) lives behind the same `opacity-0 group-hover:opacity-100` overlay, so they're all invisible until hover — and there's no hover state on a touch screen at all. Fixing just the spinner would still leave every other action unreachable on mobile.

I'll open `fix/git-panel-mobile-actions` that removes the hover-gating at the root across `GitPanel.svelte` and `GitLog.svelte` (branch/commit/remote/stash rows, both nested-repo and flat views), so actions are always visible instead of conditionally un-hidden — same fix shape you found, applied everywhere it exists. I'll credit this PR in the replacement's description.

Reopen here anytime if a narrower fix turns out to be preferable — happy to talk it through.

PR #526, opened 2026-09-08
feat(settings): add a project usage tab to Settings -> Device

srinoryanti:
## Summary
Adds a Project Usage tab to Settings → Device listing live CPU, RAM and storage
for every project in one place — searchable, filterable by status and sortable
— and stabilizes the device cards in the tab beside it so the panel opens
instantly and reports steady numbers.

## Why
Per-project resource figures only existed in the Project Info modal, reachable
one project at a time from the project navigator, so there was no way to see
where the machine's CPU, RAM and disk actually went. The Device tab itself also
showed a long skeleton while other tabs rendered instantly, then disagreed with
itself on each open: the title flipped between "This Device" and "Server",
Battery/Swap/Storage cards appeared and vanished, and CPU was inflated by its
own probe storm.

## Changes

**Per-project usage:**
- `backend/ws/projects/overview.ts` (new) — `projects:overview` returns every
  project the caller can see with the same CPU/RAM/storage figures
  `projects:info` reports. Scoped to the connection: `getAll()` for admins,
  `getAllForUser()` otherwise, with the snapshot cache keyed per scope.
- `backend/ws/projects/info.ts` — `peekFolderStats()` serves last-known folder
  stats and queues a refresh behind a four-slot pool, so the overview never
  waits on a walk and the descriptor fan-out stays bounded.
- `frontend/.../AboutDeviceSettings.svelte` — a segmented control splits the
  hardware cards from the project list so only one is mounted at a time, and
  the 3s heartbeat refreshes the visible tab only. The list is one dense row
  per project (name over path, storage over CPU/RAM) with status pills for
  Running/Idle, ordering by size, name or last opened, search across all
  projects, and `measuring` / `unavailable` storage states. Summary cards above
  the list always describe every project, never the filtered subset.
- `frontend/.../device-types.ts` (new) — wire types shared between the module
  script's prefetch caches and the component.

**Device cards:**
- `backend/ws/system/device-info.ts` — two-tier probing (fast OS-delta CPU,
  memory and network per request; battery, disk and GPU cached 60s and
  refreshed in background), sticky battery presence, OS network fallback,
  1500ms response cache.
- `frontend/.../SettingsModal.svelte` — prefetch both endpoints on modal open
  for admins, joined rather than duplicated by the panel's own mount, so the
  first click on either tab lands on data.

## Security impact
`projects:overview` is a list endpoint over data with a per-user ACL. A
project's name and absolute path are not public, and the Device tab's admin
gate is client-side only, so the handler resolves its project set from the
connection rather than listing the table. Non-admins see only their assigned
projects; the per-scope cache keeps an admin snapshot from reaching a member.

## Test plan
- `bun run check` (0 errors) / `bun run lint` / `bun run test` (1183 pass) green
  on this branch merged with current `main`
- `bun test backend/ws/projects/overview.test.ts` — 6 pass, covering storage
  totals by state and footprint ordering
- Open Settings → Device as an admin: cards render from prefetch; switch to
  Project Usage and the list is already there, filling in as walks land, with
  storage reading "Measuring…" until one completes
- Switch tabs back and forth: only the visible tab issues requests, and the
  one being revealed refreshes immediately rather than after a tick
- Filter to Running / Idle and switch ordering: the pill counts and the
  "N of M" badge track the list, the summary cards stay unchanged
- Sign in as a non-admin and call `projects:overview` directly: only assigned
  projects come back
- Start/stop a project shell and watch its dot flip within a tick
- Compare Memory/Disk/Network/Battery against Task Manager, Explorer,
  `Win32_Battery`

## Screenshots
Before/after of both tabs (light and dark).

## Notes
- Opening Settings as an admin warms both endpoints even if Device is never
  visited. The overview build itself is cheap (process table only), but it does
  queue folder walks; they are bounded at four concurrent and the storage cache
  backs off up to 5 minutes for expensive folders.
- The project list is one column by design — the settings pane is ~580px wide
  and a second column leaves the path too narrow to identify a project by.
- `ProjectInfoModal.svelte` is unchanged from `main`; the truncated-scan
  disclosure stays there and the new rows mirror it as a `≥` prefix.
- Runtime QA in a live app has not been done.

### DIFF
--- a/backend/ws/projects/index.ts
+++ b/backend/ws/projects/index.ts
@@ -7,6 +7,7 @@
  * - crud.ts: HTTP endpoints for CRUD operations (list, create, get, update, delete)
  * - status.ts: Real-time status updates and watching (get-status, watch, unwatch, events)
  * - presence.ts: User presence management (update-presence with broadcast)
+ * - overview.ts: Read-only recap of idle projects (projects:overview)
  */
 
 import { createRouter } from '$shared/utils/ws-server';
@@ -14,6 +15,7 @@ import { crudHandler } from './crud';
 import { statusHandler } from './status';
 import { presenceHandler } from './presence';
 import { infoHandler } from './info';
+import { overviewHandler } from './overview';
 
 export const projectsRouter = createRouter()
 	// CRUD Operations (HTTP)
@@ -26,4 +28,7 @@ export const projectsRouter = createRouter()
 	.merge(presenceHandler)
 
 	// Per-project resource info (storage + process-isolated CPU/RAM)
-	.merge(infoHandler);
+	.merge(infoHandler)
+
+	// Read-only recap of idle projects (counts + summed storage + top 5)
+	.merge(overviewHandler);
--- a/backend/ws/projects/info.ts
+++ b/backend/ws/projects/info.ts
@@ -22,10 +22,12 @@ import { collectProcessTree, getHostFacts, getProcessTable } from '../../host/me
 
 /** Ceiling on entries visited by one walk, so a monorepo cannot pin a core. */
 const MAX_ENTRIES = 300_000;
-/** Wall-clock ceiling — the backstop for slow disks and network mounts. */
-const WALK_DEADLINE_MS = 15_000;
+/** Wall-clock ceiling — the backstop for slow disks and network mounts.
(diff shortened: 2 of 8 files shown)


ArgaFairuz:
a

